PR #89, opened 2026-08-18
FLB#87 | Use fishing-method defaults and species quick-select chips

### LINKED ISSUE #87: [UX] Use fishing-method defaults and species quick-select chips
### Context
The Catch details flow now supports optional species, weight, length, method, bait/lure, notes and caught date/time. The current edit UI still asks the angler to type species and fishing method manually even though the profile/onboarding flow already captures preferred fishing choices.

This creates unnecessary friction. An angler who usually fly fishes for Brown Trout should not have to repeatedly type `Fly` and `Brown Trout` for most catches.

The current Profile model does **not** yet capture measurement-unit preferences. It has display/profile fields plus `PreferredFishingTypes` and `PreferredSpecies`, but no weight or length unit settings. #87 must add these preferences so Catch weight/length entry can be shown in the units the angler actually uses.

The profile currently uses the term **Fishing Type**, while Catch uses **Fishing Method**. Standardise the user-facing concept on **Fishing Method**. `Fly`, `Spinning`, `Bait`, etc. read naturally as methods. Broader taxonomy such as Sea/Coarse/Game can be revisited later without blocking this UX improvement.

### Goal
Use the angler&#39;s profile/onboarding preferences to make Catch recording/enrichment substantially faster:

- the user can choose preferred fishing methods
- one method can be their default
- preferred species can be associated with a method
- one species can be the default for a method
- the user chooses preferred **weight unit** and **length unit**
- Record Catch and Catch Edit use those unit preferences for measurement entry/display
- Record Catch surfaces the likely method/species as immediately tappable pills/chips
- the first catch of a session starts from profile defaults; subsequent catches should reuse the last method/species selected for that local fishing session/`Record another catch` flow without silently changing profile defaults
- profile/onboarding uses chip-style multi-select rather than comma-separated free text where appropriate
- unusual catches remain possible through a searchable `More…`/other path

The profile should provide intelligent defaults and shortlists; it must never restrict what the angler can record.

### Product terminology
Use **Fishing Method** consistently in user-facing UI for this work.

Examples:

```text
Method
[ Fly ✓ ] [ Spinning ] [ Bait ] [ More… ]

Species
[ Brown Trout ✓ ] [ Rainbow Trout ] [ Sea Trout ] [ Salmon ]
[ More… ]
```

Profile/onboarding is multi-select:

```text
Preferred methods
[ Fly × ] [ Spinning × ]

Preferred species for Fly
[ Brown Trout × ] [ Sea Trout × ] [ Salmon × ]
```

Catch selection is single-select.

### Measurement-unit preferences
Capture unit preferences on the user&#39;s profile/onboarding so weight and length entry are natural for the angler.

At minimum support:

```text
Weight unit
- Kilograms (kg)
- Pounds (lb)

Length unit
- Centimetres (cm)
- Inches (in)
```

Use typed enums/value objects/constants rather than arbitrary strings where consistent with repository conventions.

The profile preferences control **display and entry units**, not canonical storage. Preserve a single canonical Catch representation in the domain/database (currently #19 uses kilograms for `Weight` and centimetres for `Length`). Convert at the UI/application boundary as appropriate so changing a user&#39;s unit preference never changes the physical measurement stored for an existing Catch.

Examples:

```text
Profile:
Weight unit = lb
Length unit = in

Catch UI:
Weight [ 4.5 ] lb
Length [ 18 ] in

Persisted canonical values:
Weight = converted kilograms
Length = converted centimetres
```

Changing profile preference later must only change how values are displayed/entered; it must not rewrite historical Catch measurements.

Do not infer the permanent preference solely from locale. Onboarding may propose a sensible initial choice, but the user must be able to choose and edit it explicitly.

### Domain/data modelling principles
Do **not** model species as belonging exclusively to one fishing method. A species can be caught by multiple methods.

Use a canonical species catalogue and a canonical fishing-method catalogue, then store user preferences separately.

Conceptually:

```text
Species
- Id
- Name
- ...

FishingMethod
- Id
- Name
- ...

UserFishingMethodPreference
- UserId
- FishingMethodId
- IsDefault

UserFishingSpeciesPreference
- UserId
- FishingMethodId
- SpeciesId
- IsDefault

Profile/User preferences
- PreferredWeightUnit
- PreferredLengthUnit
```

Exact schema names should follow repository conventions, but keep the separation of canonical catalogue data from user preference data.

### Session/default behaviour
Profile defaults are long-term preferences; session choices are temporary context.

```text
Profile default = what I usually do
Session default = what I am doing right now
Catch value     = what happened for this fish
```

For the first Catch after starting the Record Catch flow, use the profile default method/species when those fields are empty.

If the angler changes method/species and taps `Record another catch`, carry the last selected method/species forward for the next Catch in that local session. Do **not** silently update their profile defaults.

Do not carry weight or length forward between catches.

A full FishingSession domain model is not required by this ticket unless the existing architecture already has one. Prefer the smallest local/session-state mechanism that preserves offline behaviour and can later migrate into a formal session/trip feature.

### Behaviour
For a user whose defaults are:

```text
Default method: Fly
Preferred Fly species:
- Brown Trout (default)
- Rainbow Trout
- Brook Trout
- Sea Trout
- Salmon

Weight unit: lb
Length unit: in
```

Record Catch / Catch Edit should surface:

```text
Method
[ Fly ✓ ] [ other preferred/recent methods ] [ More… ]

Species
[ Brown Trout ✓ ] [ Rainbow Trout ] [ Brook Trout ] [ Sea Trout ] [ Salmon ]
[ More… ]
```

Selecting a different method should update the preferred species shortlist for that method.

Do not show an unbounded wall of chips. Show a sensible short list, prioritised by:

1. explicit default
2. explicit preferred choices
3. optionally recent/frequent choices later, if implemented separately

Do not silently rewrite explicit profile defaults based on usage statistics in this issue.

### Chip / pill interaction
Use pills/chips because the likely choices should be visible without opening a dropdown.

#### Profile/onboarding
- multi-select chips
- searchable typeahead input
- selecting a canonical result creates a chip
- chips can be removed easily
- user can mark one preferred method as default
- within a method, user can mark one preferred species as default

Do not retain comma-separated species entry as the primary interaction once the chip experience exists.

#### Catch edit/recording
- single-select chips
- default choice selected initially when the Catch field is empty
- user can tap a different visible chip in one action
- `More…` opens searchable selection from the full catalogue
- existing saved Catch values must display even if they are not currently in the user&#39;s preferred shortlist

### Canonical species catalogue
Introduce/extend a canonical species catalogue suitable for later use by Catch stats, personal bests, search and leaderboards.

Do not duplicate species per method.

Seed enough common species/methods to make the feature useful, but keep seed data maintainable. Do not attempt a complete global fish taxonomy in this ticket.

The implementation should preserve the ability to expand the catalogue later.

### Onboarding/profile integration
The onboarding journey should capture enough information to make these defaults useful from the first Catch:

- preferred fishing methods
- default fishing method
- preferred species per method
- default species per method
- preferred weight unit (`kg` / `lb`)
- preferred length unit (`cm` / `in`)

Profile must allow the user to edit these later.

Existing profile visibility/privacy toggles remain unchanged unless directly needed for the field rename.

### Catch integration
Use the preferences in both the fast Record Catch flow and the Catch details UI introduced by #19.

Record Catch should show Method and Species quick-select chips because those values are likely known at the waterside and can usually be accepted in one tap/no taps. Do **not** add weight/length as required or carried-forward fields on the fast Record Catch page.

When a Catch already has a Species/Method value, do not overwrite it merely because the user&#39;s default changed later.

Defaults apply only when the relevant Catch field is empty/new.

Weight/length fields in Catch Edit must label/display/parse using the user&#39;s preferred units while preserving canonical kg/cm storage.

Saving still follows the existing offline-first #19/#17 architecture:

```text
select method/species
→ save same Catch Id locally
→ metadata becomes pending
→ best-effort sync
```

Do not create a parallel Catch save/sync path.

### Acceptance Criteria
- [ ] User-facing terminology is standardised on `Fishing Method` for this workflow.
- [ ] Canonical FishingMethod catalogue exists.
- [ ] Canonical Species catalogue exists.
- [ ] Species are not duplicated per fishing method.
- [ ] User can select multiple preferred methods.
- [ ] User can choose one default method.
- [ ] User can select preferred species for a method.
- [ ] User can choose one default species per method.
- [ ] User can choose preferred weight unit (`kg` or `lb`).
- [ ] User can choose preferred length unit (`cm` or `in`).
- [ ] Unit preferences persist on the profile and are editable later.
- [ ] Profile/onboarding uses searchable chip/pill entry rather than comma-separated text as the primary UX.
- [ ] Record Catch shows likely Method choices as immediately tappable single-select pills.
- [ ] Record Catch shows likely Species choices as immediately tappable single-select pills.
- [ ] First Catch uses profile defaults when method/species are empty.
- [ ] `Record another catch` carries the last selected method/species forward for the local session.
- [ ] Weight and length are not carried forward between catches.
- [ ] Catch edit shows likely Method choices as immediately tappable single-select pills.
- [ ] Catch edit shows likely Species choices as immediately tappable single-select pills.
- [ ] Explicit default appears first/selected when the Catch field is empty.
- [ ] Changing method updates the species shortlist for that method.
- [ ] `More…` or equivalent gives access to the full searchable catalogue.
- [ ] An existing Catch value is preserved and shown even if it is not in the current preferred shortlist.
- [ ] Defaults do not overwrite already-entered Catch data.
- [ ] Catch weight/length display and entry use the user&#39;s preferred units.
- [ ] Catch persistence remains canonical kg/cm regardless of display preference.
- [ ] Changing profile units does not rewrite historical Catch measurements.
- [ ] Same Catch Id/offline-first sync behaviour from #17/#19 is preserved.
- [ ] English and French strings are provided.
- [ ] Automated tests cover preference persistence, unit conversion, defaults, session carry-forward, chip selection and Catch integration.

### Scope restrictions
Do NOT implement:

- automatic species recognition from photographs
- a complete worldwide species taxonomy
- silent machine-learning/profile default changes from catch history
- Catch statistics/personal bests
- FishingVenue association
- guide/client delegated recording
- a new Catch synchronisation architecture
- broad redesign of unrelated profile fields
- a full FishingSession/trip domain feature solely to carry method/species forward
- changing canonical Catch measurement storage away from kg/cm in this ticket
- Terraform/infrastructure changes unless a migration requires normal application DB deployment only

### Implementation planning note
Before implementation, inspect current `main` after #19 and the current profile/onboarding model. Add an exact file-by-file Implementation Plan to this issue before handing it to Cursor, following the workflow now used for FishingLogBook issues.

### Dependencies
- #19 Catch details

### Definition of Done
- [ ] Fishing Method terminology consistent
- [ ] canonical method/species catalogues implemented
- [ ] profile/onboarding preference/default model implemented
- [ ] weight/length unit preferences implemented
- [ ] chip/pill UX implemented on Record Catch and Catch Edit
- [ ] session method/species carry-forward implemented without changing profile defaults
- [ ] Catch edit converts preferred display units to/from canonical kg/cm safely
- [ ] offline/sync behaviour preserved
- [ ] targeted and full required tests green
- [ ] EN/FR localisation present
- [ ] no unrelated scope added

## Implementation Plan (file-by-file)

1. `src/FishingLogBook.Domain/Profiles/Profile.cs`
   - Extend the profile entity with canonical default preferences: preferred methods, default method, per-method preferred species/default species, and preferred measurement units.
2. `src/FishingLogBook.Shared/Dtos/ProfileDto.cs`
   - Add the same user preference fields to the shared profile contract returned from the API.
3. `src/FishingLogBook.Shared/Dtos/UpdateProfileDto.cs`
   - Extend the update contract so profile edits can persist default method/species and unit preferences.
4. `src/FishingLogBook.Application/Args/UpdateProfileArgs.cs`
   - Carry the new preference values through the application layer update request.
5. `src/FishingLogBook.Application/Profiles/Services/ProfileService.cs`
   - Preserve the existing profile update flow while applying defaults, unit preferences, and list sanitisation without mutating historical catches.
6. `src/FishingLogBook.Infrastructure/Persistence/ProfileRepository.cs`
   - Persist the new preference columns and keep backward compatibility with the existing `PreferredFishingTypes` / `PreferredSpecies` data model.
7. `src/FishingLogBook.Db.Migrations/01_Tables/...`
   - Add the required profile migration(s) for weight/length unit preferences and any default method/species columns needed by the feature.
8. `src/FishingLogBook.Web/Features/Profile/Pages/Profile/Profile.razor` and `Profile.razor.cs`
   - Replace the free-text species entry with searchable multi-select chip UX, plus explicit default-method/default-species and unit preference controls.
9. `src/FishingLogBook.Web/Features/Catch/Models/CatchModel.cs`
   - Preserve the local session carry-forward data while keeping the canonical kg/cm storage model untouched in the stored catch data.
10. `src/FishingLogBook.Web/Features/Catch/Pages/RecordCatch/RecordCatch.razor` and `RecordCatch.razor.cs`
   - Populate method/species quick-select chips from profile defaults and the current session, without overwriting existing catch values.
11. `src/FishingLogBook.Web/Features/Catch/Pages/CatchEdit/CatchEdit.razor` and `CatchEdit.razor.cs`
   - Add the single-select pill UI, default selection logic, display conversion to the user’s preferred units, and a `More…` searchable catalogue fallback.
12. `src/FishingLogBook.Web/Features/Catch/Offline/*`
   - Keep the same offline-first catch save/sync model and ensure method/species carry-forward remains local-session only.
13. `src/FishingLogBook.Web/Localization/UiStrings.resx` and `UiStrings.fr.resx`
   - Add all new EN/FR strings and replace remaining Fishing Type copy with Fishing Method terminology.
14. `tests/FishingLogBook.Web.Tests/...`
   - Add bUnit coverage for preference persistence, canonical catalogue defaults, unit conversion, chip selection, session carry-forward, and catch integration.
15. `dotnet test` / repo validation
   - Run the targeted suites first, then the required full validation pass before completing the PR.

This plan intentionally keeps scope to #87 and leaves unrelated profile or catch features unchanged.

kingkeamo:
Closes #87.

Uses the angler's profile preferences to make catch recording faster: canonical fishing-method and species catalogues, per-user preferences with one default each, weight/length unit preferences, and single-select quick-chips on Record Catch and Catch Edit.

Later rounds added a mobile compression pass on Record Catch, a calm application error state, a crash fix for navigation-time cancellation, and cache-first preference loading. Those are summarised under **Review rounds** below.

> **Pending push:** the final commit `8c2c893` (ownership/freshness hardening from the `b5f782a` review) is local only — SSH push is not available from the agent session. The PR head is `b5f782a` until it is pushed.

## What this delivers

**Catalogues and preferences**
- Canonical `FishingMethod` and `Species` tables; species are not duplicated per method.
- `UserFishingMethodPreference` / `UserFishingSpeciesPreference` hold the per-user selection, with partial unique indexes enforcing at most one default method, and one default species per method.
- New CQRS slice under `Application/FishingPreferences/` behind `GET /api/fishing-catalogue`, `GET|PUT /api/profiles/me/fishing-preferences`.

**Measurement units**
- `PreferredWeightUnit` / `PreferredLengthUnit` on the profile, editable from the Profile page.
- Catch Edit labels, parses and displays in the angler's units while **storage stays canonical kg/cm**.
- Conversion lives in one place (`Features/Catch/Services/MeasurementService`) using exact factors (1 lb = 0.45359237 kg, 1 in = 2.54 cm) and `MidpointRounding.AwayFromZero`.
- **No-drift rule:** on save, if the displayed value is unchanged, the stored canonical value is kept byte-identical. Proven by tests that open/save five times in lb and in inches.

**Chips**
- Record Catch and Catch Edit show single-select method/species chips, default first and pre-selected when the field is empty.
- Changing method reshapes the species shortlist for that method.
- A saved value that is not in the shortlist is still shown and stays selected.
- `More…` opens a searchable full-catalogue picker (`Common/Modals/CataloguePicker`).
- Profile uses multi-select chips with explicit default-method and default-species-per-method controls, plus the searchable picker for adding species.

**Session carry-forward**
- Carry-forward belongs to the explicit *Record another catch* flow only, as component-local state. Starting Record Catch fresh — including navigating away and back — uses profile defaults. Profile defaults are never silently changed.
- The carried species also carries whether it was explicitly chosen, so a value that was only auto-defaulted last time stays defaultable.
- Weight and length are deliberately not carried forward (Record Catch has no measurement fields at all).

**Offline preference cache**
- Preferences and the catalogue are cached per angler in a **separate** `FishingLogBookPreferences` IndexedDB database. The Catch database schema and version are untouched (still version 3), enforced by schema-safety tests.
- `AnglerPreferencesProvider` reads that cache first, renders from it, and refreshes from the API in the background, so Record Catch does not wait on the network.

## Migrations

Three new expand-only scripts, all idempotent:

- `01_Tables/202608181900_87_CreateFishingCatalogue.sql`
- `01_Tables/202608181901_87_AddProfileUnitPreferences.sql`
- `02_SeedData/202608181902_87_SeedFishingCatalogue.sql` (5 methods, 12 species)

**Manual post-merge step:** apply migrations with
`dotnet run --project src/FishingLogBook.Db.Migrations.App -- --run`.

No destructive changes. The legacy `Profile."PreferredFishingTypes"` / `"PreferredSpecies"` columns are kept; retiring them is tracked in #90.

**Terraform: not changed, nothing applied.**

## Security and privacy

- `UserId` always comes from the request-scoped `ICurrentUser`; the client never supplies it. An API test posts a raw JSON body containing a foreign `userId` and asserts the save happens for the authenticated angler.
- All three endpoints `RequireAuthorization()` and return 401 when identity is unresolved.
- Unknown catalogue ids return 400; infrastructure failures return 503 (matching `ProfileEndpoints`), so a database outage is not reported as a client error.
- The Profile page cannot reach catch storage — asserted by an architecture test — so changing a unit preference cannot rewrite historical catch measurements.
- The preference cache is keyed on the owner id resolved from server identity. A background refresh re-resolves the owner **after** the API call and discards the result if the signed-in account changed, so one angler's data can never be written under another's key.
- The error boundary never renders exception messages, stack traces or internal URLs, and invents no correlation id.

## Tests

**1061 .NET tests, 161 JS, 35 browser (1 pre-existing skip) — all green.**

| Project | Coverage added |
|---|---|
| `Application.Tests` | `FishingPreferenceService` (4 files), update command + validator, both query handlers + validator, profile unit mapping and persistence |
| `Infrastructure.Tests` | Real PostgreSQL via Testcontainers: catalogue reads, preference reads, replace (insert/update/delete), both FK violations, both partial-unique default indexes, transaction rollback, cross-user isolation, profile unit round-trip |
| `Api.Tests` | All three endpoints through the real mediator chain: unauthenticated, validation failure, unknown catalogue entry, 503, success, client-supplied-userId rejection |
| `Shared.Tests` | Serialisation of the five new DTOs and both enums, including absent-unit defaulting |
| `Web.Tests` | `MeasurementService` (30 tests incl. no-drift), `AnglerPreferencesProvider` (15, incl. cache-first, account-switch-during-refresh, failed-persist freshness), `FishingPreferenceClient` HTTP contract, `CataloguePickerModal`, `AppErrorBoundary` state and recovery actions, MainLayout chrome-on-page-failure and synchronisation cancellation, `OfflineOperation` cancellation, and chip/default/carry-forward/unit behaviour on all three pages with French copy assertions |
| JS (vitest) | Owner-scoped preference store, and schema-safety tests proving preference storage cannot modify or share the Catch database schema |

### Round 2 — `9cabfa3`

**1. Catch Edit did not apply defaults to empty fields.** Added `ApplyProfileDefaultsToEmptyFields()` after binding, plus default-species-on-method-change when species is empty. Non-empty stored values always win.

**2. Record Catch could overwrite an explicit species.** Added explicit-vs-defaulted tracking — chip taps and `More…` choices mark the value explicit, and `SelectMethod` only replaces a species that was auto-defaulted.

**3. Carry-forward was broader than the ticket.** `ICatchSessionService` deleted; carry-forward is now component-local state, written at save and read only by `RecordAnotherCatch()`.

**4. Legacy `PreferredSpecies` could still be wiped.** The page captures the legacy list once at load as a frozen baseline and writes back `baseline ∪ chip species`.

**5. Fishing Type terminology.** The `#profile-show-fishing-types` toggle and its strings are removed rather than relabelled; a real Method-based visibility control lands with #90.

**7. Seed IDs.** Replaced with 17 generated UUIDs; `Code` remains the semantic identity.

**8. Contract cleanup task opened: #90.**

### Round 3 — the two Record Catch edge cases

Both fixed and covered: switching to a method with no default species now clears an auto-defaulted species, and every method-change path (chip, `More…`, and formerly typing) runs the same default-species logic while preserving an explicitly chosen species.

### UX pass — Record Catch mobile compression

- Duplicate free-text Method and Species inputs removed; chips plus `More…` are the only selectors, with the explicit-vs-auto-default logic preserved.
- Header compressed: subtitle removed, `h4` → `h6`, reduced container and stack spacing.
- `Take photo` is the primary of the two side-by-side photo actions; `Choose photo` is secondary.
- Location section compressed to a compact `Location on · Private` status, keeping the "exact fishing spot is private by default" message.
- A loading spinner covers the page until the angler preferences have loaded, matching the existing `CatchEdit` / `CatchList` / `Profile` pattern.

**Consequence worth flagging:** with the free-text fields gone, off-catalogue method/species values can no longer be entered on Record Catch. Saving is unaffected (only a photograph is required) and Catch Edit still has free-text fields, so a value can be added afterwards.

### UX pass — application error state

- `AppErrorBoundary` replaces the filled red `MudAlert` with a calm centred state: restrained outline error icon, **Something went wrong**, a short non-technical explanation, primary **Try again** (`ErrorBoundary.Recover()`, not a force reload) and secondary **Go home**. `role="alert"`, `aria-hidden` on the decorative icon, EN + FR.
- A second boundary inside `MainLayout` wraps `@Body`, so a page failure keeps the app bar, drawer and navigation; the `App.razor` boundary remains the last resort.
- The stock `#blazor-error-ui` bar — the surface an exception that escapes the renderer actually reaches — is restyled to match. Its text is static pre-Blazor HTML and therefore remains English; see the deliberate gaps below.
- Exception logging and the diagnostics path are unchanged. `LoggingErrorBoundary` was not touched.

### Crash fix — navigation-time cancellation

Navigating to a page while a background synchronisation was in flight could take the whole application down. Disposing the previous page cancelled its token, the offline write failed with `TaskCanceledException`, and the handler meant to absorb that failure read `.Token` from an **already disposed** `CancellationTokenSource`, so an `ObjectDisposedException` escaped into the renderer. A reload worked because there was no previous page to dispose.

- Synchronisation wrappers capture the token before the `try`.
- Every failure-reporting call uses `CancellationToken.None`, so reporting cannot fail for the same reason as the operation it reports.
- `OfflineOperation` rethrows caller cancellation without emitting a failure diagnostic; cancellation is not a failure.
- Cancellation catches are filtered on the caller's own token, so an unrelated `TaskCanceledException` (an HTTP timeout, for example) is still logged rather than swallowed.
- `CatchList` ignores synchronisation updates once disposed.

### Performance — Record Catch load

Opening Record Catch made five serial awaits before rendering: resolve owner, then profile, then catalogue, then preferences, then the location prompt. The three preference reads are independent and now run together, and the catalogue load runs alongside the location prompt. `Profile` had the same serial pattern on load and was given the same treatment.

The provider also remembers the loaded preferences for the session, keyed by owner id, so returning to Record Catch or Catch Edit does no work at all. Profile hands its saved values straight to the provider, which remembers them in memory before persisting them best-effort — a failed cache write cannot leave the next Record Catch serving an older default. The API read is bounded to ten seconds.

## Self review

- Issue/AC: PASS — all 29 Acceptance Criteria implemented and covered by tests.
- Architecture/CQRS: PASS — endpoint → mediator → handler → service → repository; handlers take `I*Service` only; transactions live in the repository.
- Rules: PASS — `cqrs.md` Mapster violation reverted; handler naming, one-type-per-file DTOs, block bodies, and no-comment rules restored.
- Security/privacy: PASS — see above.
- Database/migrations: PASS — expand-only, idempotent, split per `database.md`, exercised for real by the Testcontainers suite. The Catch IndexedDB schema and version are unchanged.
- Tests/coverage: PASS. The two hardening tests from the latest review were verified non-vacuous by removing each fix in turn and confirming the test fails.
- Browser: N/A — the Playwright suite drives a static JS harness, not an authenticated Blazor host, and its README already records authenticated Profile UI as out of scope. Forcing coverage would require test-only authentication in `src/`, which `testing-blazor.md` forbids on a product ticket. Residual risk is real-browser rendering of the chips, the picker dialog and the compressed Record Catch layout; that is compensated by bUnit component tests, API tests, and Testcontainers repository tests. A testing-infrastructure follow-up is warranted if authenticated browser coverage is wanted.
- Web/UI/localisation: PASS — EN and FR both 205 keys, verified key-for-key identical; no orphaned keys introduced (`Catch_RecordSubtitle` and `Action_Reload` were removed when their last usage went).
- Code smells: PASS after fixes — conversion factors and profile length limits extracted to constants, chip shortlist logic shared in `CatchChipOptionModel`, duplicated validation moved from the service into the validator, and `Invalidate()` replaced by `SetAsync(...)` so preference persistence has one owner.
- CI/deployment: PASS — migration apply is the only manual step; no Terraform, config, secret or CI changes.

### Findings discovered during self-review

1. `FishingPreferenceEndpoints.ToDataResult` took `ValidatedResponse` directly, which the `DependencyInjectionTests` endpoint scan correctly flagged as an unresolvable handler parameter.
2. Profile length limits (100 / 200 / 50) were duplicated between the server validator and the Web page.
3. No test proved that changing unit preferences cannot rewrite stored catches.
4. `AppErrorBoundary.GoHome()` would not recover the boundary when the failing page *was* the home page, because no `LocationChanged` fires for a same-route navigation.
5. `.blazor-error-boundary::after` carried hard-coded English copy.
6. Four existing location tests asserted only `"Location on"`, which still passes as a prefix of the new copy.
7. `#blazor-error-ui` forced `color-scheme: light only`, painting a light bar over a dark app.
8. Bumping the Catch IndexedDB version for the preference store caused a `VersionError` on downgrade and a 5s `OfflineDbReadTimedOut` on upgrade during local testing.
9. `TypeAdapterConfig.GlobalSettings` is process-wide and eight Application test classes mutated it in parallel, producing a ~1-in-4 flake.

### Fixes made

1. `ToDataResult<TResponse, TData>` constrained to `ValidatedResponse`.
2. Added `Shared/Constants/ProfileDetailConstants` and used it in both places.
3. Added an architecture test asserting the Profile page injects no catch store or catch client.
4. `GoHome()` recovers the boundary explicitly before navigating.
5. Removed the `::after` rule — every boundary supplies `ErrorContent`, so it never rendered.
6. Strengthened those assertions to the full `Location on · Private` / `Localisation activée · Privée`.
7. Removed the forced colour scheme so the bar follows the theme.
8. Moved the preference cache into its own `FishingLogBookPreferences` database and reverted the Catch database to version 3, with schema-safety tests to keep them apart.
9. Added `MapsterTestConfig` (lock + once-only `Scan`) and documented the pattern in `testing-csharp.md` so other projects reusing these rules do not hit it.

### Remaining deliberate gaps

- Profile save is two API calls (profile, then preferences) and is not atomic; a failure between them leaves preferences unsaved and surfaces one generic error.
- Legacy `PreferredFishingTypes` is preserved and still governs the public profile, but is no longer editable from the Profile page — it is the old "fishing type" taxonomy, which #87 explicitly defers. A Fishing Method-based public visibility control lands with #90.
- Legacy `PreferredSpecies` values cannot be removed through the new chip UI during the expand phase; #90 migrates them into the catalogue model.
- Off-catalogue method/species values can no longer be typed on Record Catch (see the UX pass above); Catch Edit still allows them.
- A background preference refresh updates the cache and the in-memory copy but does not rebind an already-rendered page; the next page load gets the fresh values.
- Remembered preferences are held for the session, so a preference change made on another device is not picked up until the app is reloaded. A same-device Profile save updates it immediately.
- `#blazor-error-ui` copy is static pre-Blazor HTML and cannot use the resource files, so it remains English. The `AppErrorBoundary` state — which is what ordinary page failures reach — is fully localised.
- No per-component inline error states were added. The `MainLayout` boundary already keeps the chrome and replaces only the content area; adding inline states to individual pages would have meant redesigning screens this pass was told not to touch.
- **Unrelated to #87, worth its own ticket:** during local testing the three preference endpoints each took ~46 s to return `200`, finishing within ~170 ms of each other — they were blocked on one shared resource, not three slow queries. Cache-first hides this on Record Catch, but a cold cache or a new device still pays it. Not investigated in this PR.

## Follow-ups

1. **Contract phase — #90 (opened)** — backfill the legacy profile arrays into the catalogue model, stop writing them, then drop the columns in a later release. Required by the expand/contract rule in `database.md`.
2. **Browser test infrastructure** — an authenticated Blazor host for Playwright, if authenticated browser coverage is wanted. Not yet opened.
3. **API cold-response investigation** — the ~46 s first-response behaviour described above. Not yet opened.
4. **`Features/*/Services` tidy-up** — `OfflineOperation` (a suffix-less static helper used cross-feature), `CorrelationDelegatingHandler` (an HTTP handler) and `DiagnosticSynchroniser` (sync belongs in `Offline/` per `blazor.md`) are misplaced. Agreed to defer to its own task.

## Note

`dbb9902` on this branch adds `.mcp.json` (Rider MCP config) and is unrelated to #87.


kingkeamo (commented):
Strict #87 review — **not ready to merge yet**. The schema/CQRS/unit-conversion work is substantially better than the earlier branch, but I found four behavioural/data blockers plus a few cleanup items.

## BLOCKERS

1. **Catch Edit does not apply defaults when the stored fields are empty.**

`CatchEdit.LoadAsync` loads preferences and then `BindFormAsync` simply assigns `catchRecord.Method ?? string.Empty` / `SpeciesName ?? string.Empty`. There is no default application afterwards. `MethodOptions` puts the default first, but selection is based on equality with `_method`, so an empty Catch has no selected method; `SpeciesOptions` is then empty because there is no method preference to resolve.

This directly misses #87: **“default choice selected initially when the Catch field is empty”** / **“Explicit default appears first/selected when the Catch field is empty.”** Add regression coverage for (a) both fields empty, and (b) method present + species empty. Existing non-empty values must still win.

2. **Record Catch can overwrite explicitly entered species with a method default.**

`RecordCatch.SelectMethod` always assigns the new method's default species when one exists. If the angler has already typed/tapped an explicit species and then changes the method, that explicit Catch value is silently replaced. #87 explicitly says defaults only populate empty/new fields and **must not overwrite already-entered Catch data**.

Please distinguish default-derived state from an explicit user choice. Switching method may replace a species that was only auto-defaulted, but must preserve a species the angler explicitly selected/typed. Add a regression test for explicit species -> change method -> species preserved, while retaining the existing default-to-default method-switch behaviour.

3. **Session carry-forward is broader than the ticket and defeats first-catch profile defaults.**

`CatchSession.Remember(method, species)` runs after every successful Save, and `ICatchSessionService` is scoped and never cleared. A later navigation back to Record Catch therefore restores the previous catch even when the angler did **not** tap `Record another catch`. The PR explicitly advertises this cross-navigation behaviour, but #87 says: first Catch after starting Record Catch uses profile defaults; **if the angler taps `Record another catch`**, carry the last method/species into the next Catch.

Please make carry-forward belong to that explicit flow/session boundary rather than every successful save/app lifetime. This also avoids stale/user-crossing volatile state. Add a test proving a fresh Record Catch entry that was not reached via `Record another catch` uses profile defaults, while `Record another catch` still carries method/species only.

4. **Legacy `PreferredSpecies` can still be destructively wiped.**

`PreferredSpeciesNames()` preserves `_preferredSpecies` only while `_selectedMethods.Count == 0`. For an existing user with legacy `PreferredSpecies = ["Pike", "Tench"]` and no new preference rows, simply selecting a method (with no species yet) changes the next profile save to `PreferredSpecies = []`. Likewise beginning to configure the new model can silently drop legacy values the new UI never showed them.

That contradicts #87's backward-compatibility/expand requirement and the PR claim that no existing user loses data. The existing test only covers the zero-selected-method case, so it misses this path. Preserve legacy data safely through the expand phase (or implement an explicit safe migration strategy) and add regression coverage for legacy species + newly selected method/no new species.

## SHOULD FIX BEFORE MERGE

5. **Fishing Type terminology is still visible in this workflow.** `Profile_ShowFishingTypes` still renders “Show preferred fishing types to other anglers” / the French equivalent. #87 and its implementation plan explicitly require user-facing terminology to be standardised on **Fishing Method** and to replace remaining Fishing Type copy. Because that toggle currently governs legacy data, please reconcile the backing behaviour as well — do not just relabel stale Fishing Type data as methods.

7. **Seed IDs are still synthetic patterned UUIDs.** We explicitly agreed during schema review to use fixed, generated UUIDs for production catalogue rows, with `Code` as the semantic identity. The current `a7f1c2d0-0000-...0001` / `b3e4a5c0-0000-...` values still look like fixture IDs. Please replace them before this migration is applied anywhere permanent (and update the `.http` examples accordingly).

8. Repo `database.md` requires the expand/contract cleanup task to be **actually opened before merge**. The PR currently says “Follow-ups to open before merge”, but I cannot find that contract issue yet. Please open/link it before we merge; browser-test-infrastructure can remain a separate follow-up if desired.

## What looks good in this pass

- relational `FishingMethod` / `Species` catalogues are independent;
- per-user method preferences and per-method species preferences are modelled correctly;
- composite FK prevents species prefs outside a selected method;
- partial unique indexes enforce max one default method and max one default species per method;
- no global `DefaultSpecies`/`DefaultFishingMethod` profile shortcut remains;
- unit preferences are profile-level and DB constrained;
- Catch canonical kg/cm storage remains intact;
- no-drift conversion approach is sensible;
- owner identity for preference APIs comes from `ICurrentUser` rather than caller data;
- same Catch/offline sync path appears preserved;
- no `Shared.Helpers` dumping-ground remains.

Please address the blockers, add the missing regressions, run the targeted/full suites again, and push a new head. I will re-review the complete updated diff. CI for the current head is still running at the time of this review.

kingkeamo (commented):
Round 3 strict review on `9cabfa3`.

The previous four blockers are materially improved, the legacy-data preservation is now safe for the expand phase, the broad `ICatchSessionService` has been removed correctly, the seed UUIDs are fixed, Fishing Type UI copy is removed, and #90 exists. `.mcp.json` is intentional repo tooling and should remain.

I still see two blocking #87 behaviour gaps:

1. **Offline profile defaults / units are not preserved.** `RecordCatch.LoadCatalogueAsync()` and `CatchEdit.LoadPreferencesAsync()` source the catalogue/preferences/profile only through API clients. When those calls fail offline, Record Catch falls back to no profile defaults/chips and Catch Edit falls back to `Kg`/`Cm`. #87's ACs are not online-only: first Catch must use profile defaults, Catch Edit must use the user's preferred units, and the existing offline-first flow must be preserved. A previously-configured angler going waterside offline should not lose those preferences. Please persist/cache the minimum user-scoped catalogue/preferences/unit state needed for offline Record/Edit (using the existing offline architecture rather than inventing another catch path), refresh it when online, fail closed across account changes, and add regression coverage for offline use plus user isolation.

2. **Auto-default species is still not distinguished from an explicit Catch value in Catch Edit, and stale auto-defaults can survive method changes.** With an empty Catch, `ApplyProfileDefaultsToEmptyFields()` produces e.g. Fly + Brown Trout. If the angler then taps Spinning, `SelectMethod()` sees a non-empty species and preserves Brown Trout, even though Brown Trout was auto-defaulted rather than entered/saved by the angler. That means the Spinning default (e.g. Pike) is not selected. Record Catch has explicit-vs-defaulted tracking now, but also leaves the previous auto-default species in place if the newly selected method has no default species. Please track whether the current species is explicit/saved vs auto-defaulted in both flows: method changes may replace/clear only an auto-defaulted species, while an existing saved/typed/chip/More choice must remain untouched. Add tests for (a) empty Catch -> Fly/Brown default -> switch to Spinning -> Pike, (b) auto-default -> switch to method with no default -> species clears, and (c) explicit/saved species survives a method change.

One non-blocking contract note: Profile currently renders every catalogue method as a chip and only Species has the searchable picker. #87 describes searchable chip/typeahead entry for Profile/onboarding. With five seeded methods this is usable, but please either add the same bounded/searchable `More…` path for methods now or explicitly track it as a follow-up if you consider the current bounded catalogue sufficient for #87.

CI for this head is still running, so this review does not treat the PR's local green claim as final CI evidence.

kingkeamo:
Round 2 re-review on head `9cabfa3`: the four original blockers are materially addressed, #90 is open, the seed IDs are cleaned up, and `.mcp.json` is intentionally retained for Rider/Claude MCP.

Two Record Catch edge cases still need fixing before I’d call #89 clear:

1. **Switching from an auto-defaulted species to a method with no default species leaves the old species behind.** `SelectMethod` only assigns `_selectedSpecies` when `defaultSpecies is not null`. If the current species was auto-defaulted (for example Brown Trout under Fly) and the user changes to a method that has no default species, the old Brown Trout should be cleared/defaultable rather than silently carried onto the new method. The current code leaves it in place. Please add a regression test for this case.

2. **Typing a method bypasses the method-change/default-species behaviour.** The free-text field now binds through `MethodInput`, but that setter only changes `_selectedMethod` / `_methodIsExplicit`; it never applies the same species logic as `SelectMethod`. Typing a known preferred method such as `Spinning` therefore changes the shortlist but can leave the previous auto-defaulted species (for example Brown Trout) selected instead of applying Spinning’s default (Pike). Since free text is a supported input path, method changes should behave consistently whether they come from a chip, `More…`, or typing, while still preserving an explicitly chosen species. Please cover the typed-known-method path with a test too.

Everything else from the previous review looks substantially improved. CI for `9cabfa3` is still in progress, so this is code-review status only, not a final green-light yet.

kingkeamo:
### UX follow-up from local testing

The #87 functionality is good, but the Record Catch screen needs a mobile compression pass before merge. Please make these UX changes while preserving the existing default/species behaviour and offline-first save flow:

1. **Remove the duplicate Method and Species text inputs from Record Catch.** The chips are the primary input; `More…` provides catalogue/search selection. Having a selected chip and a text box containing the same value wastes substantial mobile vertical space and creates two interaction paths for the same choice.
2. **Keep Method/Species selection tap-first.** Method: preferred chips + `More…`; Species: preferred chips for the selected method + `More…`. Preserve the explicit-vs-auto-default logic already fixed in review. Do not regress method changes/default-species behaviour.
3. **Compress the page header.** Remove the `Photograph, then save on this device` subtitle and reduce the vertical spacing/size around `Record catch`. The target is a compact mobile page header rather than a large desktop-style hero heading.
4. **Make camera the primary photo action.** `Take photo` should have stronger prominence/full-width treatment; `Choose photo` can be a secondary action rather than two equal-weight side-by-side buttons. Preserve existing camera/gallery functionality.
5. **Compress the location section without weakening the privacy message.** The current multi-line explanation + two large actions consumes too much vertical space. Keep the fact that the exact fishing spot is private by default, but present permission/status much more compactly. Once location permission/status is established, prefer a compact status such as `Location on · Private` rather than repeatedly occupying several lines.
6. **Mobile acceptance target:** on a normal phone at initial Record Catch load, the Method choices, Species choices, and primary `Take photo` action should be visible without scrolling where practical. This is a waterside quick-entry screen; vertical real estate matters.

This is intentionally a compression/refinement of the existing UI, not a redesign. The current chips/default interaction is good. Please keep the existing localisation approach and add/update tests where the changed interaction requires it.

kingkeamo:
### Add to the #87 UX refinement: application error state

Please include a focused tidy-up of the global/application error boundary encountered during local testing, alongside the Record Catch mobile compression work already requested.

Keep the error boundary and the existing exception logging/diagnostics. This is presentation and recovery work — do **not** suppress, swallow, or mask exceptions.

The current full-content red error screen is too aggressive and looks like a developer/fatal crash screen. Replace it with a calm, mobile-friendly product error state:

- keep the normal application/background styling; do not paint the entire content area red;
- use a small warning/error icon with restrained error colouring;
- short heading such as **Something went wrong**;
- short non-technical explanation such as **We couldn't load this screen. Try again, or return home.**;
- primary **Try again** action;
- secondary **Go home** action;
- retain normal application chrome/navigation where it is safe to do so;
- never show exception messages, stack traces, internal URLs, or other technical details to the angler;
- continue routing the underlying exception through the existing logging/diagnostic path;
- only show an opaque correlation/reference ID if one already safely exists — do not invent a telemetry system for this UX change;
- do not make generic claims such as **Your catch is safe** unless the application actually knows local persistence completed;
- where a recoverable component-level failure can be handled locally, prefer a compact inline error + retry rather than unnecessarily replacing the whole page; the global boundary remains the last-resort fallback;
- preserve localisation and accessibility conventions used by the app;
- add/update bUnit coverage for the user-visible boundary state and its recovery actions.

Desired feel: normal app background, compact centred error content, clear recovery actions, and a small restrained error accent — something that looks intentionally designed rather than the application having turned red and died.

Please treat this and the previously posted Record Catch feedback as one UX compression/refinement pass. Do not redesign unrelated screens or change the #87 defaults/offline behaviour while doing it.

kingkeamo (commented):
Re-review of the new cache-first Record Catch path (`b5f782a`). The performance direction is good: Record Catch reads the owner-scoped preference cache first and refreshes in the background, and the normal Profile-save path writes the freshly saved profile/preferences into that cache before invalidating the provider. That means a normal same-device Profile change should be visible immediately on the next Record Catch without waiting for the API.

Two edge cases still need hardening:

1. **[privacy / ownership] Re-check the owner before caching a background API refresh.** `GetAsync()` resolves `userId`, then `RefreshAsync(userId)` can wait on `_loadLock` and later call the API using the current authentication context. If the signed-in account changes while that refresh is waiting/in flight, fresh data for the new account can potentially be written under the old `userId` cache key. Before `TryCacheAsync(userId, fresh, ...)` and before remembering the refreshed value, resolve the current owner again and only persist/apply it if it still equals the original `userId`. Please add the account-switch-during-refresh regression test, not just the existing remembered-value isolation test.

2. **[freshness] A successful Profile save can still leave Record Catch showing stale defaults if the local cache write fails.** `CacheForOfflineUseAsync` deliberately swallows cache failures, then `AnglerPreferences.Invalidate()` clears memory. If an older cache entry exists, the next Record Catch returns that stale entry immediately and its background refresh does not rebind the already-rendered page. Prefer updating the provider's in-memory current value directly from the successfully saved `ProfileDto` + `FishingPreferencesDto` (and best-effort persisting it), or otherwise ensure invalidation after a successful Profile save cannot make the next Record Catch serve an older cache value. Add a regression test for: old cached default -> Profile save changes default -> cache persistence failure -> next provider/Record read still gets the newly saved default.

The cache-first strategy itself should stay; this is about ownership/freshness around it, not reverting to API-first and making Record Catch slow again.

kingkeamo:
Both points from the `b5f782a` re-review are fixed in `8c2c893`. Cache-first is untouched.

### 1. Re-check the owner before caching a background refresh

Confirmed — `RefreshAsync` resolved the owner once, in `GetAsync`, and then made the API call later, so an account switch while it waited on `_loadLock` or sat in flight could write the new account's data under the previous angler's key. It now re-resolves after the API returns and discards the result on a mismatch:

```csharp
var fresh = await TryLoadFromApiAsync(CancellationToken.None);
if (fresh is null) { return; }

var currentUserId = await ResolveOwnerAsync(CancellationToken.None);
if (currentUserId != userId) { return; }

await TryCacheAsync(userId, fresh, CancellationToken.None);
```

Regression test: `WhenTestingCacheFirst.ItShouldNotCacheABackgroundRefreshUnderTheOldOwnerWhenTheAccountChanges`. It gates `GetOwnAsync` on a `TaskCompletionSource` and waits until the refresh is provably inside the lock, switches the signed-in account, releases the API, then uses `SetAsync` as a lock barrier so the assertion cannot run before the refresh finishes. No sleeps, no polling. The existing remembered-value isolation test is kept as well.

### 2. A failed cache write could leave stale defaults

Agreed, and `Invalidate()` was the wrong primitive — clearing memory while an older cache entry survived is exactly the hole you describe. It has been removed from `IAnglerPreferencesProvider` and replaced with:

```csharp
Task SetAsync(Guid userId, AnglerPreferencesModel preferences, CancellationToken cancellationToken);
```

which remembers the value in memory **first**, then persists best-effort. Profile no longer writes the cache itself — `IAnglerPreferencesCache` is out of `Profile.razor.cs` entirely and persistence has one owner. A quota failure can no longer cost the freshly saved defaults.

Regression test: `WhenTestingSet.ItShouldServeTheSuppliedPreferencesEvenWhenPersistingThemFails` — stale cached `Lb` default → `SetAsync` with the newly saved `Kg`/Spinning values → cache save throws → next read returns the new values.

Two Profile page tests moved down a layer with the responsibility: they now assert `SetAsync` on the provider rather than `SaveAsync` on the cache, and the cache write itself is proven by the provider's own tests.

### Both tests verified non-vacuous

I removed each fix in turn and confirmed the matching test fails:

- guard removed → `Expected to receive no calls matching SaveAsync(11111111-…, any AnglerPreferencesModel, any CancellationToken) / Actually received 1 matching call`
- `Remember` removed from `SetAsync` → `Expected served.WeightUnit to be WeightUnitEnum.Kg, but found WeightUnitEnum.Lb`

### Validation

`dotnet format` clean · `dotnet build` succeeded · **1061 .NET tests** (Web 485) · 161 JS · 35 browser (1 pre-existing skip) — all green.

The PR description has been rewritten to cover every round rather than appending another section.

### One thing outside this PR

While diagnosing the Record Catch load time, all three preference endpoints returned `200` after **~46 s**, finishing within ~170 ms of each other — they were blocked on one shared resource rather than being three slow queries. Cache-first hides it on Record Catch, but a cold cache or a new device still pays it. I have not investigated the API side; it is listed as follow-up 3 in the description and probably wants its own ticket.

PR #95, opened 2026-08-19
FLB#92 | inject IMapper for Infrastructure catch persistence mapping

### LINKED ISSUE #92: [Technical] Standardise injected IMapper across production mapping boundaries
### Context

#87 exposed a process-wide Mapster race caused by Application code using static `.Adapt<T>()` / `TypeAdapterConfig.GlobalSettings`. That issue is being fixed in #89 by giving each container its own `TypeAdapterConfig` and using constructor-injected `IMapper` throughout the Application layer.

For scope control, #89 deliberately stops there. Existing mapping below/outside Application — particularly Infrastructure repository persistence/domain mapping — is not being converted as part of #87.

We want a dedicated cleanup so production mapping follows one predictable convention everywhere it is appropriate.

### Architectural decision

**If production code has an object-to-object adaptation responsibility, inject `IMapper` and use it consistently.**

Do not choose manual property copying merely because a mapping currently contains only two or three matching properties. A mapping that is trivial today can grow later; keeping the caller on `IMapper` lets matching-name properties continue to map by convention and lets non-trivial changes move into central Mapster configuration without rewriting every caller.

This distinction remains important:

- **Adaptation / projection between application, persistence and transport models** → injected `IMapper`.
- **Domain construction that represents behaviour/invariants/ownership/identity creation** → explicit construction; do not hide domain behaviour behind Mapster.

No production code should use static Mapster entry points or `TypeAdapterConfig.GlobalSettings`.

### Goal

Audit the remaining production solution outside the Application-layer work completed in #89 and establish one consistent mapping pattern using DI-owned Mapster configuration and injected `IMapper` wherever object adaptation is required.

This is a structural/architecture cleanup only. It must not change API contracts, persistence semantics, domain behaviour or UI behaviour.

### Scope

Audit all production projects, with particular attention to:

- `FishingLogBook.Infrastructure`
- repository row/persistence-model → Domain mapping
- Domain → persistence argument/record mapping where applicable
- any mapping in API composition/endpoints that should live behind an existing boundary
- any remaining Web/server-side C# mapping where Mapster is genuinely being used or where repeated object adaptation exists
- `FishingLogBook.DependencyInjection` registration/lifetimes needed to make the same container-owned mapper available to the layers that need it

Also search for all Mapster entry points:

- `.Adapt<T>()`
- `.Adapt(source, destination)`
- `TypeAdapter.Adapt`
- `BuildAdapter()`
- `TypeAdapterConfig.GlobalSettings`
- locally created ad-hoc Mapster configs in production code
- existing injected `IMapper`

### Required convention

Where a class adapts one model to another:

```csharp
public SomeRepository(IMapper mapper, ...)
{
    _mapper = mapper;
}

var domain = _mapper.Map<DomainType>(record);
```

Matching-name mappings should use Mapster convention. Add/update `IRegister` configuration only when mapping requires nested-path mapping, renames, ignores, conversions or other non-trivial configuration.

Do **not** hand-map an adaptation simply because it is currently small. The caller should not need rewriting when that mapping later grows from two properties to ten.

Do **not** inject `IMapper` into Domain entities/services. Domain construction remains explicit where it represents behaviour or invariants.

### Acceptance Criteria

- [ ] Audit documents every remaining production object-to-object mapping outside the Application cleanup completed in #89.
- [ ] All applicable production adaptation sites use constructor-injected `IMapper`.
- [ ] No production code uses `TypeAdapterConfig.GlobalSettings`.
- [ ] No production code uses static Mapster `.Adapt`, `TypeAdapter.Adapt` or `BuildAdapter` entry points.
- [ ] Infrastructure mapping that is mechanical projection/adaptation is moved to injected `IMapper`.
- [ ] Domain entity construction that carries ids, ownership, invariants or behaviour remains explicit and is not converted mechanically to Mapster.
- [ ] Matching-name convention mappings do not gain unnecessary empty `IRegister` configuration.
- [ ] Non-trivial mappings live in clearly named `*MappingRegistration` classes in the appropriate mapping/composition location.
- [ ] The DI-owned `TypeAdapterConfig` remains container-local and is reused by all injected mappers; no second/global config path is introduced.
- [ ] Repository/service lifetimes remain valid after mapper injection.
- [ ] Existing API/persistence behaviour is byte/semantically unchanged where applicable.
- [ ] Existing tests are updated to use isolated mapper configuration rather than shared static Mapster state.
- [ ] Architecture/regression coverage is extended so future production static Mapster usage in the newly covered layers fails CI.
- [ ] Relevant `.claude` architecture/testing rules are updated to describe the solution-wide convention rather than Application-only wording.

### Test Requirements

- Keep existing #89 Application/Api Mapster tests green.
- Add focused mapping tests only where custom mapping configuration exists; do not test Mapster's own matching-name convention unnecessarily.
- Existing Infrastructure integration tests must remain green against real PostgreSQL/Testcontainers.
- Run the full .NET suite with normal parallelisation enabled.
- Add/extend an architecture test that scans all applicable production assemblies for banned static Mapster entry points / `GlobalSettings`.
- Verify constructing multiple DI containers still produces independent `TypeAdapterConfig` instances.

### Out of Scope

- Changing Domain behaviour or entity invariants
- API contract redesign
- Database schema/migration changes
- UI redesign
- #87 fishing preference behaviour
- Replacing Mapster with another mapper
- Introducing mapping merely for object construction that is genuine domain behaviour

### Definition of Done

- [ ] Remaining production mapping audit complete
- [ ] Applicable mapping boundaries use injected `IMapper`
- [ ] No production static/global Mapster use remains
- [ ] Domain construction remains explicit where appropriate
- [ ] Architecture/rules/tests enforce the convention
- [ ] Infrastructure integration tests and full .NET suite green
- [ ] Self-review completed

kingkeamo:
Closes #92.

## Summary

Audits production mapping outside the Application-layer work done in #89 and applies the same injected-`IMapper` convention where genuine object-to-object adaptation exists.

- **`CatchRepository`** was the only Infrastructure class with hand-written adaptation: `LoadAsync` copied a private `CatchRow` persistence projection into `Catch` field-by-field. It now injects `IMapper` and calls `_mapper.Map<Catch>(catchRow)`.
- New `Infrastructure/Persistence/Mappings/CatchMappingRegistration.cs` (`IRegister`) holds that mapping, mirroring the existing `Catch → CatchDto` `.MapWith` pattern in Application. `CatchLocation.TryCreate` (which enforces domain invariants — coordinate/enum/consent validation) stays explicit, referenced from the mapping rather than hidden behind convention.
- Composition root's `TypeAdapterConfig.Scan` now scans the Infrastructure assembly too — still one container-owned config, no second/global path.
- `WhenTestingMapsterConfiguration.ProductionAssemblies` extended to cover the Infrastructure assembly, so the architecture test fails CI on any future static Mapster usage there.
- Every other Infrastructure repository (`SystemRepository`, `TestCatchRepository`, `UserIdentityRepository`, `UserPlatformCapabilityRepository`, `FishingCatalogueRepository`, `FishingPreferenceRepository`) relies on Dapper's own column-name binding straight onto the Domain type — no hand-written adaptation exists there, so nothing changed.
- `.claude/rules/cqrs.md`, `csharp.md`, `database.md`, `testing-csharp.md` updated to describe the Mapster convention as solution-wide rather than Application-only.

Full audit and plan posted to the issue before implementation: https://github.com/kingkeamo/fishing-logbook/issues/92#issuecomment-5343342151

## Deliberately out of scope

- `CatchRepository.ToRow` and the `ToParameters` helpers in `ProfileRepository`/`UserPlatformCapabilityRepository` build anonymous Dapper SQL-parameter objects (with UTC/enum conversions) — these are parameter shaping, not adaptation between two named models, so they stay manual.
- No isolated unit test for `CatchMappingRegistration` itself — `CatchRow` would need wider visibility than the mapping needs, and the existing (green) Testcontainers `CatchRepository` integration suite already asserts the exact round-trip behaviour (Location, photographs, provenance defaults, timestamps) end-to-end.

## Self Review

```
Self review:
- Issue/AC: PASS — audited every remaining production mapping site outside #89; converted the one genuine Infrastructure adaptation site to injected IMapper; kept CatchLocation.TryCreate explicit; no static Mapster entry points remain; architecture test + composition root extended to Infrastructure; rules updated to solution-wide wording.
- Architecture/CQRS: PASS — Endpoint -> IMediator -> Handler -> I*Service -> I*Repository chain unchanged; only Infrastructure repository internals and composition-root wiring changed.
- Rules: PASS — cqrs.md, csharp.md, database.md, testing-csharp.md updated; followed git-commits.md and self-review.md.
- Security/privacy: PASS — no change to auth, ownership checks, or data exposed; persisted/returned data is byte-identical.
- Database/migrations: N/A — no schema change.
- Tests/coverage: PASS — full `dotnet test` green (1092 tests across all projects), including the 31 live-Postgres CatchRepository integration tests, which now exercise the new mapping path end-to-end. Architecture test extended to scan the Infrastructure assembly. Infrastructure.Tests TestMapper and both CatchRepository test bases updated to isolated mapper configuration.
- Browser: N/A (no Web/UI changes) — skipped `npm run test:browser`; ran `npm test` (vitest), green.
- Web/UI/localisation: N/A — no Web changes.
- Code smells: PASS — one deliberate note: CatchRow.Photographs is now a mutable (non-init) property so the second query's result can ride into the mapper call; acceptable since CatchRow is an internal transport DTO with no invariants of its own.
- CI/deployment: PASS — no CI, Terraform, or migration changes required.

1. None beyond the CatchRow.Photographs mutability noted above (accepted, not a blocker).

Fixes made:
1. Injected IMapper into CatchRepository; added CatchMappingRegistration.
2. Extended composition root Mapster scan and the Mapster architecture test to cover Infrastructure.
3. Updated Infrastructure.Tests TestMapper and both CatchRepositoryTests bases to build an isolated mapper.
4. Updated .claude rules to describe the convention as solution-wide.

Remaining deliberate gaps:
- ToRow/ToParameters SQL-parameter builders left manual (Dapper parameter shaping, not model adaptation).
- No isolated mapper-only unit test for CatchMappingRegistration; covered by the extended Testcontainers integration suite instead.
```


kingkeamo (commented):
I checked this closely because the diff is smaller than #92 initially suggested. The small production diff is mostly legitimate: #89 already converted the Application layer, most Infrastructure repositories let Dapper materialise directly into their target types, and the remaining `ToParameters`/anonymous SQL bags are parameter shaping rather than named-model adaptation. So I do **not** think we should manufacture mapper usage just to make this PR larger.

I do have three #92 contract gaps to fix before merge:

1. **`CatchMappingRegistration` currently defeats the main convention #92 introduced.** It uses `MapWith(source => new Catch { ... })` and manually lists every same-named property. #92 explicitly says matching-name properties should map by convention so adding a new matching property later does not require rewriting this mapping. Keep only the genuinely non-trivial location conversion in configuration (and any property that truly cannot map conventionally); let `Id`, `UserId`, provenance, timestamps, species, weight/length, method, notes, photographs, etc. map by convention. The current `MapWith` is essentially the hand-written mapping moved into a different file.

2. **The architecture guard does not scan the API assembly.** The pre-implementation inventory explicitly planned to add both Infrastructure and Api, and #92 asks for solution-wide protection across applicable production layers. `ProductionAssemblies` currently contains Application, Infrastructure and DependencyInjection only. Add `FishingLogBook.Api` so a future static `.Adapt`/`GlobalSettings` introduced in an endpoint/composition path fails CI as intended.

3. **#92 explicitly requires focused mapping tests where custom mapping configuration exists.** We now have a custom `CatchRow -> Catch` registration because `Location` requires `CatchLocation.TryCreate`, but the PR deliberately adds no focused mapping test. Please add a small test around the custom registration, focused on the non-trivial location conversion (not Mapster's ordinary same-name behaviour). Keep the existing Testcontainers tests as the end-to-end persistence protection too.

Apart from those points, the size of the PR no longer worries me: the audit supports why there aren't many more production conversions, and we should not invent persistence DTOs or inject mapping where Dapper already performs direct materialisation.

kingkeamo (commented):
One additional cleanup to include in this PR, following the review discussion:

**Replace anonymous SQL parameter objects returned as `object` with named, strongly typed persistence parameter models.**

`CatchRepository.ToRow(Catch)` currently returns `object` backed by an anonymous type. Dapper can consume that, but once returned from the method the compile-time contract is lost. For a codebase we are deliberately standardising now, prefer an internal named persistence parameter type (for example `CatchParameters` / `CatchPersistenceParameters`) and make the method return that concrete type.

Keep the actual persistence shaping explicit rather than forcing it through Mapster: this method flattens `CatchLocation` and normalises timestamps to UTC, so those transformations are meaningful persistence-boundary logic, not convention-only object adaptation.

Please also apply this as the **going-forward repository convention** in the appropriate `.claude` database/C# rule: repository helper methods should not return anonymous parameter bags as `object`; use a named internal `*Parameters`/persistence model when the parameter set is substantial or returned from a helper. Small anonymous objects created directly at a single Dapper call site are still fine where they improve readability.

While making this change, check the other repository `ToParameters` helpers identified in the #92 inventory (`ProfileRepository`, `UserPlatformCapabilityRepository`) and apply the same rule where they currently return `object`/anonymous types. Do not mechanically create types for trivial one-off inline Dapper arguments.

This is in addition to the earlier review findings: Mapster convention for same-name properties, API assembly architecture-test coverage, and focused coverage for the custom Catch location mapping.

kingkeamo (commented):
One additional correction from reviewing the latest changes: please remove the new `<InternalsVisibleTo Include="FishingLogBook.Infrastructure.Tests" />` from the production Infrastructure project. We should not weaken the production assembly boundary solely so tests can directly access internal persistence implementation types. Adjust the new focused mapping coverage to test through an appropriate public production boundary (the repository/integration path is acceptable here) rather than exposing Infrastructure internals for the test.

Keep the read/write persistence distinction discussed in the existing review: the read-side persistence shape (`CatchRow`, optionally renamed to something clearer such as `CatchPersistenceRow`) is distinct from the write-side `CatchParameters`; do not merge those concepts. The important requested cleanup remains replacing helper methods that return anonymous parameter bags as `object` with named strongly typed parameter models where appropriate.

kingkeamo (commented):
Latest head addresses the earlier Mapster/configuration review well: same-name Catch properties now map by convention with only `Location` customised; API is included in the static-Mapster architecture guard; the production `InternalsVisibleTo` escape hatch is gone; and the substantial repository parameter helpers have been converted to named persistence parameter types. The DI-owned config still remains one container-local path.

I still see two blockers before #92 is complete:

1. **The focused custom-mapping regression is still missing.** #92 explicitly requires focused coverage where custom mapping configuration exists, and the earlier review asked specifically for the non-trivial `CatchLocation.TryCreate` mapping. The PR body now calls the absence of that test a deliberate gap, but that leaves an acceptance criterion and an earlier review finding unresolved. We do not need to expose `CatchPersistenceRow` or add `InternalsVisibleTo`; cover the behaviour through the public repository/Testcontainers boundary with a focused test that would fail if the custom location mapping were removed/broken (including the privacy/consent fields that `TryCreate` enforces).

2. **The legacy Catch cleanup we agreed to fold into #92 has not been done.** The current tree still contains the old `TestCatch` production slice (`Api/Endpoints/TestCatchEndpoints`, `Application/TestCatches`, `Infrastructure/Persistence/TestCatchRepository`, `Web/Features/TestCatch/...`) and `ProductionCatch` naming remains in the Catch storage/browser-test surface. We no longer need the test-vs-production Catch distinction now that the real Catch flow is established. Please audit those paths before deleting/renaming anything, remove obsolete TestCatch production scaffolding, and rename `ProductionCatch` terminology to the normal `Catch` vocabulary where it is only transition-era naming. Preserve durable storage/API compatibility where a literal persisted key/schema name cannot safely change; document any such deliberate compatibility exception rather than blindly renaming it.

Because that second item touches Web/offline storage, update the affected unit/JS/browser coverage and run the relevant browser suite as part of the final validation rather than treating browser coverage as N/A for the completed #92 scope.

CI on current head `4f90fc7` is still in progress, so this is not ready for final dev/device testing yet.

Issue #97, opened 2026-08-19
[UX] Turn Catch list into the angler Logbook experience


kingkeamo:
## Context

The core Catch journey is now functional: offline-first recording, stable client-generated IDs, photographs, idempotent sync, edit details, location privacy, ownership/provenance and user fishing preferences.

The current `/catches` page was intentionally built as a functional view while those capabilities were developed. It now exposes too much implementation/state information and does not yet feel like an angler's logbook.

This ticket turns `/catches` into the primary Logbook experience without changing the underlying offline-first Catch architecture.

## Goal

An angler opening the Logbook should be able to:

1. immediately recognise catches visually;
2. scan species, date/time and useful catch details;
3. distinguish catches requiring attention from healthy catches;
4. quickly record another catch;
5. open/edit/manage a catch without a wall of action buttons;
6. find older catches using simple local search/filtering;
7. use the same experience effectively on phone, tablet and desktop;
8. continue to use the Logbook while offline.

The page should prioritise fishing information, not technical sync information.

---

## Layout direction

This is a **mobile-first photographic logbook**, not a CRUD/admin list.

### Overall page

Keep the content visually constrained. On desktop the current cards stretch across a very wide area while all useful content sits on the left. Do not simply make mobile cards 1200px wide.

Suggested direction:

```text
My Logbook                                  [ + Record catch ]
A record of your catches

[ Search catches...                         ]
[ All ] [ Fly ] [ Spinning ] [ Filters ]

Today · 3 catches

┌──────────────────────────────────────────────────────────┐
│ ┌──────────────┐  Brown Trout                      ⋮     │
│ │              │  Today · 4:50 PM                        │
│ │    PHOTO     │  Fly                                    │
│ │              │  2.1 lb  ·  43 cm                      │
│ └──────────────┘                                          │
└──────────────────────────────────────────────────────────┘

Yesterday · 2 catches
...
```

Do not copy this literally if MudBlazor/responsive constraints suggest a better implementation; it defines the intended information hierarchy.

### Mobile card

A phone card should be compact enough that several catches can be scanned without excessive scrolling.

Suggested structure:

```text
┌──────────────────────────────────────┐
│ ┌──────────┐  Brown Trout        ⋮   │
│ │          │  Today · 4:50 PM        │
│ │  PHOTO   │  Fly                    │
│ │          │  2.1 lb · 43 cm         │
│ └──────────┘                          │
│                                      │
[...]

PR #99, opened 2026-08-20
Feature/97 turn catch page into log book

kingkeamo:
Cluade crashed but it had commited all changes lest review ticket

### DIFF
--- a/src/FishingLogBook.Web/Features/Catch/Components/CatchCard/CatchCard.razor
+++ b/src/FishingLogBook.Web/Features/Catch/Components/CatchCard/CatchCard.razor
@@ -1,0 +1,104 @@
+@namespace FishingLogBook.Web.Features.Catch.Components.CatchCard
+
+<MudCard Class="catch-card" id="@($"catch-card-{Catch.Id:D}")">
+    <div class="catch-card-menu">
+        <MudMenu>
+            <ActivatorContent>
+                <MudIconButton Icon="@Icons.Material.Filled.MoreVert"
+                               aria-label="@Loc["Catch_MoreActions"]"
+                               OnClick="@context.ToggleAsync"
+                               id="@($"catch-card-menu-{Catch.Id:D}")" />
+            </ActivatorContent>
+            <ChildContent>
+                <MudMenuItem Icon="@Icons.Material.Filled.Edit"
+                             Href="@EditHref"
+                             id="@($"catch-card-edit-{Catch.Id:D}")">
+                    @Loc["Catch_Edit"]
+                </MudMenuItem>
+                @if (Catch.Location is not null)
+                {
+                    <MudMenuItem Icon="@Icons.Material.Filled.LocationOn"
+                                 OnClick="LocationPrivacyAsync"
+                                 id="@($"catch-card-location-privacy-{Catch.Id:D}")">
+                        @Loc["Catch_ManageLocationPrivacy"]
+                    </MudMenuItem>
+                }
+            </ChildContent>
+        </MudMenu>
+    </div>
+
+    <a class="catch-card-body" href="@EditHref" id="@($"catch-card-link-{Catch.Id:D}")">
+        <div class="catch-card-media">
+            @if (_thumbnailUrl is not null)
+            {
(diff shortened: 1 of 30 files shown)


kingkeamo:
Review of current head `ee49d935` against #97: the visual/structural direction is good, but this is not ready for dev/device testing yet.

I found four blockers:

1. **CI is red and the parent CatchList test fixture was not updated for the new dependencies.** `CatchList` now injects `ITimeService`, `IAnglerPreferencesProvider`, `ICatchDateGroupingService` and `ILoggingService`, but `BaseCatchListTest` on this branch is byte-for-byte the same as `main` and registers none of them. Please fix the failing suite rather than only the new component tests.

2. **The ticket-required CatchList behaviour coverage is largely missing.** The existing `CatchListTests` files are unchanged from `main`, while #97 explicitly requires coverage for the integration behaviour introduced in `CatchList`: newest-first + local date grouping, search by species/method/bait, filters + clear, no-match vs no-catches vs load-failure states, unexpected-load-error logging, preferred measurement units, and `StateChanged` refresh. `CatchCard`/`CatchFilters` unit tests are useful, but they cannot prove the page wires those behaviours together correctly. Please update/add focused CatchList tests for the new page logic.

3. **Cancellation is currently treated as an unexpected load failure.** `LoadAsync()` catches every `Exception`, sets `_loadFailed`, clears the list, and calls `LogErrorAsync`. Its cancellation token is the component-owned CTS that `Dispose()` cancels. A normal navigation/disposal while owner/store/time/preferences work is in flight can therefore be logged as an application error. We already established elsewhere in this app that caller/component cancellation is not a failure. Filter `OperationCanceledException` for this component's token before the general exception path and cover it with a regression test.

4. **#97 explicitly asks for responsive/browser coverage where the existing browser-test conventions support it, and this PR adds none.** This ticket materially changes phone/desktop layout (`md=6`, constrained container, action menu, photographic cards). Please add the existing-project-style browser checks for phone + desktop, no horizontal overflow, reachable primary/menu actions, and cards not becoming excessively wide/empty. If a specific browser requirement truly cannot be exercised by the current harness, document exactly which one and why, but the whole responsive acceptance area should not be left unvalidated.

The underlying UX decisions look aligned with #97: local-first store remains in place, healthy sync is quiet, pending/failed states stay recoverable, self/self provenance is removed, card/menu components are split coherently, and EN/FR resources were updated. Please finish the integration/test/error-handling work above and get CI green before final device review.

kingkeamo (commented):
This is a strong first pass at #97 and the overall direction is right: the MudCard/MudGrid layout, compact actions, local search/filtering, date grouping, quiet healthy sync state, unit preferences, provenance cleanup and exception logging are all good improvements.

Before we merge, please finish the following UX requirements/clarifications:

1. **Show Bait/Lure on the card when populated.** There is enough space in the current card and this is useful logbook information. Keep it compact and omit the row entirely when empty.

2. **Show a short Notes preview when populated.** Keep this deliberately bounded (e.g. approximately 1–2 lines with ellipsis/clamping) so a long note cannot make one card enormous. Do not render an empty Notes row.

3. **Multiple photographs must be discoverable/viewable from the Logbook card.** The current implementation always takes `Catch.Photographs.FirstOrDefault()` and therefore silently hides the rest. Use the MudBlazor 9.8.0 components/APIs available in this repo to provide a compact carousel/gallery treatment when a catch has multiple photographs. One photograph should remain visually simple; navigation/dots/arrows/swipe affordance should only appear where useful. Preserve offline rendering and do not change R2/photo persistence architecture.

4. **Do not add photograph editing to #97.** Adding/removing/reordering photographs on an existing catch has offline sync/server/R2 lifecycle implications and will be a separate ticket. This PR only needs to display all photographs already attached to the catch.

5. Please add focused tests for Bait/Lure present/absent, Notes present/absent and bounded preview behaviour, one photograph, multiple photographs, and no photograph. Multiple-photo coverage should prove the additional photos are actually reachable/displayed rather than only proving the first image renders.

Please also complete the remaining #97 validation/self-review that was interrupted when Claude crashed, and update the PR description with the final implementation/test results.

Issue #100, opened 2026-08-20
Sync catches down from server into the Logbook list


kingkeamo:
## Context

The Logbook (`/catches`, issue #97) reads exclusively from local IndexedDB (`ICatchStore`). There is no server endpoint or client path that lists a user's catches, so a catch recorded on one device (or before local storage was cleared) never appears on another device/browser profile. `ICatchClient` only has `UpsertAsync`/photo-upload/location-visibility methods; `CatchEndpoints.cs` only has a single-catch-by-id `GET`. This ticket adds the missing read path: list a user's catches from the server and merge them into what the Logbook displays, alongside the existing local-first records.

## Goal

An angler opening the Logbook on a device that does not yet have all their catches in local storage should still see every catch they own, with photographs rendered from the server via a URL (not downloaded into IndexedDB in this ticket).

## Requirement references

- `docs/Requirements.md`, `docs/Architecture.md` (offline-first Catch architecture)
- Issue #97 (Logbook redesign) — this ticket extends it with a read-from-server path; #97 explicitly kept Logbook local-first and out of scope for changing sync protocol, so this is deliberately a separate ticket.

## Acceptance Criteria

1. `GET /api/catches` (authorized) returns all catches owned by the current user (`CatchViewDto`-shaped), each including per-photograph metadata (`Id`, `ContentType`, `Url`) where `Url` is a presigned R2 download URL built the same way `ProfileService` builds `PhotographUrl` (`IObjectStorage.CreateDownloadUrlAsync` keyed by the photo's `ObjectKey`).
2. `ICatchService.GetMyAsync` / `ICatchRepository.GetByUserIdAsync` follow the existing `GetCatch`/`GetByIdAsync` two-query pattern (catch row + photograph rows), ordered newest-first by `CaughtOn`.
3. Web `ICatchClient` gains `GetAllAsync(CancellationToken)` calling `GET api/catches`.
4. `CatchList.LoadAsync()` merges server results with local `ICatchStore.GetAllAsync` results: local records win on `Id` collision; server-only records display with `SyncStatus.Synchronised` and their photo rendered from the presigned `Url` (no local bytes, no IndexedDB write in this ticket).
5. `CatchCard` renders a photograph from `RemoteUrl` when local `Bytes` are absent, falling back to the existing placeholder only when neither is present.
[...]

PR #102, opened 2026-08-20
FLB#20 | Catch Details/Edit page with photograph management

### LINKED ISSUE #20: [Feature] Complete Catch Details/Edit experience with photograph management
## Context

The Catch journey is now close to a usable MVP:

- #14–#19 established production Catch recording, offline persistence, location/privacy, synchronisation, provenance and later metadata editing;
- #87 established user fishing preferences/defaults;
- #97 turns `/catches` into the primary photographic Logbook with chronological grouping, local search/filtering, compact cards and multi-photo viewing.

The remaining gap is the individual Catch experience.

The current edit page is functional but is not yet a polished mobile-first Catch Details/Edit experience. It also cannot manage photographs after the Catch has been recorded.

For MVP we want one coherent place where an angler can open a Catch from the Logbook, see all of its useful information, edit that information, manage its photographs, and manage location/privacy without navigating through a collection of disconnected screens.

This ticket completes the original `docs/Requirements.md` #8 Catch History requirement that an individual Catch shows complete details.

## Goal

When an angler opens a Catch from the Logbook, they should get a complete, mobile-first Catch Details/Edit experience where they can:

1. see all existing Catch information;
2. edit the same Catch without changing its stable client-generated `Id`;
3. view all existing photographs;
4. add additional photographs;
5. remove individual photographs;
6. edit species, method, measurements, bait/lure, notes and caught date/time;
7. manage location/privacy through the existing privacy model;
8. save while offline;
9. have all metadata/photo changes synchronise correctly when online or after reconnect;
10. return naturally to the Logbook.

This is the final major Catch UX pass before Home/onboarding/beta-readiness work.

---

## UX direction — one Catch Details/Edit page

Do **not** create a separate read-only detail page plus a disconnected edit page unless the current architecture proves that is necessary.

Preferred MVP direction:

```text
/catches/{id}/edit
```

becomes the complete Catch Details/Edit experience.

Opening a Catch from the Logbook should land here.

The user should not have to choose between "View" and "Edit" for ordinary use. The page should be pleasant to read when untouched and straightforward to edit.

## Mobile-first collapsible sections

The page must not become one very long wall of form controls.

Use logical collapsible/expandable sections so the angler can scan the Catch and open only the area they want to work with.

Strongly prefer MudBlazor's installed expansion-panel/accordion-style components where appropriate. Inspect the MudBlazor version actually installed in the repository before selecting APIs.

Suggested information architecture:

```text
Brown Trout
Today · 08:08

[ PHOTOGRAPH CAROUSEL / GALLERY ]

▼ Catch details
  Species
  Fishing Method
  Bait / Lure
  Caught date/time

▼ Measurements
  Weight
  Length

▼ Photographs
  existing photographs
  add photograph
  remove photograph

▼ Location & privacy
  location captured / not captured
  current privacy level
  manage privacy

▼ Notes
  full notes

[ Save changes ]
```

Exact grouping can be refined after inspecting the current page, but keep sections based on user concepts rather than implementation layers.

### Expansion behaviour

On phone:
- sections should minimise vertical overload;
- the most useful section(s) may start expanded;
- the user can quickly expand/collapse with touch;
- section headers should provide useful summary content where practical.

On tablet/desktop:
- use the additional width deliberately;
- sections may be wider or selectively open by default;
- do not simply stretch a narrow phone form across the entire application shell.

This interaction pattern should be implemented cleanly enough that the Profile page can later adopt the same concept, but **do not refactor Profile in this ticket**.

Do not build a generic accordion framework prematurely. Use MudBlazor and small feature-specific components where useful.

---

## MudBlazor-first implementation

Use the installed MudBlazor components for behaviour/structure rather than recreating them with custom JavaScript.

Inspect the exact installed version before implementation.

Likely useful components to evaluate:

- expansion panels / accordion components for logical detail groups;
- `MudCarousel` for multiple photographs;
- `MudImage` for image rendering;
- `MudFileUpload` or the existing Record Catch capture/upload component/path where appropriate;
- `MudTextField` / numeric controls for editable values;
- `MudChip`/`MudChipSet` for preferred species/method choices where the existing edit implementation already uses them;
- `MudButton` / `MudIconButton` for Add/Remove/Save;
- `MudAlert` only where status/error messaging actually requires attention;
- `MudSkeleton`/progress components where existing loading behaviour warrants them.

Reuse the interaction patterns from the Record Catch flow where they are already proven.

Do not introduce another UI library.

---

## Catch information

The details experience should present/edit the current Catch fields where available:

- Species
- Fishing Method
- Weight
- Length
- Bait / Lure
- Notes
- Caught date/time
- Photographs
- Location/privacy summary
- meaningful provenance where the angler/recorder differs from the current user

Do not show raw internal identifiers or technical storage/sync fields in the normal UI.

Healthy synchronisation should remain visually quiet.

Pending/failed sync may be shown when it requires user attention, consistent with #97.

Use the user's current preferred measurement units for display/edit presentation while preserving the canonical persisted measurement representation.

Reuse the canonical Fishing Method/Species preference/catalogue work already present. Do not reintroduce legacy free-text preference infrastructure.

---

## Photograph management — MVP scope

This ticket deliberately includes photograph management for an **existing Catch**.

The MVP requirement is intentionally small:

### View

- show every photograph already attached to the Catch;
- use a swipeable/touch-friendly carousel/gallery consistent with #97;
- a single photograph should remain visually simple;
- no-photo state must be deliberate and clear.

### Add

Allow the owner to add one or more photographs to the existing Catch.

Reuse the existing Record Catch photograph capture/selection pipeline as much as possible:

- camera capture on supported mobile devices;
- file/photo selection where supported;
- existing image validation/compression/storage conventions;
- stable client-generated photograph IDs;
- local blob storage before network work.

Adding a photograph must work offline.

### Remove

Allow the owner to remove an individual photograph.

Removal must work offline and persist across close/reopen.

Do not implement photograph reordering for MVP.

Do not implement image editing/cropping/filtering.

Do not implement a separate media library.

---

## Photograph lifecycle / synchronisation

Photograph removal is not only a visual operation. Implement the smallest correct end-to-end lifecycle.

### Local-first boundary

For add/remove operations:

```text
user action
→ update local Catch/photo IndexedDB state
→ UI confirms local save
→ synchronisation runs best-effort if online
→ otherwise operation remains pending until reconnect
```

Network failure must not undo a successful local action.

### Adding photographs

New photo:
- stable photograph ID generated client-side;
- blob stored locally;
- photograph state starts pending;
- existing Catch ID remains unchanged;
- existing synchronised photographs remain untouched;
- #17 photograph upload flow is reused/generalised where already appropriate;
- retry remains idempotent.

### Removing photographs

A photograph deleted locally must not simply reappear after the next synchronisation.

The sync model needs an explicit durable concept for pending deletion/tombstone or another equally safe mechanism.

Server-side removal must:
- verify Catch ownership using server-side current user identity;
- remove the appropriate `CatchPhotograph` persistence record;
- never permit deletion from another user's Catch;
- be idempotent when retried.

### R2 cleanup

When a synchronised photograph is removed, the corresponding R2 object should eventually be deleted so abandoned private objects are not retained forever.

Use the existing `IObjectStorage` boundary rather than exposing R2 credentials to Web.

R2 deletion should be:
- server-side only;
- based on the persisted object key, not a client-trusted arbitrary key;
- idempotent;
- safe to retry.

Do not make the user wait for R2 deletion before the local photograph disappears from their Catch.

If DB removal succeeds but R2 deletion temporarily fails, do not restore the photograph to the user's Catch. Log the failure with safe context and retain/retry cleanup using the smallest architecture-consistent mechanism already available. If no suitable retry mechanism exists, stop during the implementation plan and document the smallest safe choice before adding new infrastructure.

Do not make the bucket public.
Do not proxy full photograph bytes through the API.
Do not expose R2 credentials or presigned URLs in logs.

---

## Offline behaviour — non-negotiable

The complete page must remain usable offline for locally available Catch data.

Offline the owner must be able to:

- open an existing locally stored Catch;
- see its locally stored photographs;
- edit metadata;
- add photographs;
- remove photographs;
- save edits;
- close/reopen and see those edits;
- see appropriate pending-sync state.

The local Catch ID remains the permanent identity.

Do not replace local-first editing with a server-first CRUD page.

Do not require connectivity merely to open Catch Details.

---

## Save model

Use the same semantics established by #19:

```text
edit Catch
→ validate
→ write local Catch/photo state
→ local Save succeeds
→ mark changed metadata/photos appropriately pending
→ trigger existing synchroniser best-effort
```

If metadata does not change, do not reset metadata sync unnecessarily.

If photographs do not change, do not reset/re-upload them.

If only photographs change, do not unnecessarily dirty unrelated Catch metadata.

If a photograph is removed, preserve the deletion state until server/R2 cleanup is confirmed as required by the chosen state model.

Keep race protection for edits while an older sync operation is in flight.

---

## Location & privacy

Reuse the existing location/privacy architecture.

The Catch Details page should show a concise summary such as:

```text
Location captured
Private
```

or equivalent localised presentation.

Use the existing `LocationPrivacyModal` / `IModalService` approach to change privacy.

Do not show precise coordinates unnecessarily in the default Details UI.

Do not change the existing privacy semantics in this ticket.

Do not capture a new location automatically when merely editing other Catch details.

---

## Validation

Retain/reuse the existing #19 validation rules for:

- measurements;
- text lengths;
- caught date/time;
- future date protection.

Client validation improves UX but server validation remains authoritative.

Photograph add validation must reuse existing Record Catch limits/types rather than introducing a second inconsistent rule set.

All user-visible validation/error strings must be EN/FR localised.

---

## Error handling / privacy

Do not silently swallow unexpected exceptions.

Any touched/new `try/catch` must log the actual exception at the appropriate level before presenting generic safe UI messaging.

Do not log:

- photograph bytes;
- precise private coordinates;
- presigned URLs;
- object-storage credentials;
- authentication tokens;
- sensitive request payloads.

Server storage failures should include safe IDs/context where useful, not private media/location data.

---

## Component structure

Do not allow the new page to become a monolithic `.razor` file.

Inspect current Catch feature conventions and #97 components.

Likely feature components might include concepts such as:

```text
Features/Catch/Pages/CatchEdit/

Features/Catch/Components/CatchPhotoManager/
Features/Catch/Components/CatchDetailsSection/
```

Names/structure should follow current conventions after inventory.

Do not create generic abstractions solely because Profile may use expansion panels later.

Keep feature-specific code under the Catch feature.

---

## Navigation

The #97 Logbook card should continue opening this page.

Provide a clear return/back path to the Logbook.

After Save:
- do not strand the user;
- communicate successful local save clearly but unobtrusively;
- remain on the Catch or return to Logbook according to the cleanest existing navigation convention;
- do not claim "Synchronised" merely because the local save succeeded.

Avoid accidental loss of unsaved form changes.

If practical with current patterns, warn before navigating away from genuinely dirty unsaved changes. Do not introduce a global navigation-state framework solely for this ticket.

---

## Accessibility

- expansion panel headers keyboard accessible;
- photograph carousel usable by keyboard/touch where MudBlazor supports it;
- meaningful localised image alt text;
- remove-photo controls have explicit accessible labels;
- destructive photo removal requires a clear user action and must not rely only on an icon with no accessible name;
- form labels accessible;
- validation associated with fields;
- focus moves sensibly after add/remove/modal actions;
- no colour-only sync/error state.

---

## Testing

Before implementation, inspect existing #19, #17, Record Catch, photo-store, synchroniser, R2/API and #97 tests.

Add/update focused coverage proving at minimum:

### Details/Edit UI

1. owner can open a Catch from local storage offline;
2. all current metadata fields display;
3. sections render with the intended collapsible structure;
4. user can edit every #19 editable field;
5. preferred measurement units are used;
6. full notes are available in Details/Edit;
7. meaningful provenance appears only where required;
8. location/privacy summary is correct;
9. Location Privacy modal still works;
10. Save updates the same Catch ID locally.

### Photographs

11. no-photo state;
12. one-photo state;
13. multiple photographs are viewable/reachable;
14. existing local photographs render offline;
15. user can add a photograph offline;
16. added photo survives reload;
17. adding photo preserves existing photo IDs/states;
18. user can remove an unsynchronised photograph locally;
19. user can remove a synchronised photograph locally;
20. removed photo stays removed after reload;
21. removed photo does not reappear after sync;
22. removing one photo does not remove/re-upload unrelated photos;
23. photo deletion retries are idempotent;
24. another user cannot delete a photograph from the Catch;
25. server photograph record is removed correctly;
26. R2 cleanup uses the server-known persisted object key;
27. R2 cleanup never trusts arbitrary object keys from Web;
28. R2 deletion failure is logged safely and does not resurrect the Catch photograph;
29. new photograph upload continues to use the existing presigned/direct-to-R2 model;
30. photograph bytes are not routed through the API.

### Offline/sync

31. metadata-only edit does not re-upload photos;
32. photo-only edit does not unnecessarily reset metadata sync;
33. local Save succeeds while network is unavailable;
34. pending changes synchronise after reconnect;
35. existing in-flight metadata race protection still works;
36. provenance/owner/location are not mutated by ordinary editing;
37. synchronisation retries remain idempotent.

### UX/localisation

38. EN/FR strings for section headers/actions/errors/photo management;
39. phone layout does not horizontally overflow;
40. desktop layout remains intentionally constrained/readable;
41. expansion interactions work using existing bUnit/browser-test conventions where appropriate.

Run targeted tests during implementation, then the complete repository-required validation before PR creation.

---

## Implementation inventory required before coding

Before changing production code, perform one focused inventory against current `main` after #97 is merged.

Inspect specifically:

- `CatchEdit` current page/component/tests;
- `CatchList` / `CatchCard` navigation;
- Record Catch photograph capture/preview/removal implementation;
- `CatchModel` / `CatchPhotographModel`;
- IndexedDB Catch/photo stores;
- `CatchSynchroniser` photograph state machine;
- existing photograph upload URL / metadata recording flow;
- `CatchPhotograph` DB schema/repository;
- `IObjectStorage` and R2 implementation;
- current API/application photograph operations;
- `LocationPrivacyModal` / `IModalService`;
- measurement/preferences providers;
- localisation;
- MudBlazor installed version and suitable expansion/upload/carousel components;
- applicable `.claude` rules.

Add a concise implementation plan to this issue before coding that names:

- exact files/components to create/change;
- proposed collapsible section structure;
- MudBlazor components to use;
- how Record Catch photo UI/code will be reused;
- local photo-add state flow;
- local photo-delete/tombstone flow;
- server photograph-delete path;
- R2 cleanup behaviour and retry/error handling;
- tests to add/update;
- any requirement that current architecture cannot satisfy without broader scope.

Do not repeatedly rescan the whole repository.

---

## Out of scope

- photograph reordering;
- image crop/edit/filters;
- public/shared Catch detail pages;
- Catch deletion;
- personal bests;
- statistics/dashboard;
- maps;
- FishingVenue association;
- social sharing/feed/comments/likes;
- audit/history/version tables;
- changing Catch identity;
- changing owner/provenance semantics;
- changing location privacy semantics;
- Profile page redesign (future ticket/pass);
- Home/onboarding work (next phase);
- Terraform/prod infrastructure provisioning.

---

## Definition of Done

- [ ] Logbook opens a complete Catch Details/Edit experience.
- [ ] Page is mobile-first and organised into logical collapsible sections.
- [ ] All existing Catch metadata can be viewed and edited coherently.
- [ ] All existing photographs can be viewed.
- [ ] Photographs can be added to an existing Catch offline-first.
- [ ] Individual photographs can be removed offline-first.
- [ ] Removed photographs do not return after synchronisation.
- [ ] Server photograph rows are removed safely/idempotently.
- [ ] R2 objects for removed synchronised photographs are cleaned up server-side without exposing R2 credentials/keys to Web.
- [ ] No photograph reordering is introduced.
- [ ] Existing photograph upload/direct-R2 architecture is preserved.
- [ ] Metadata-only edits do not re-upload photographs.
- [ ] Photo-only edits do not unnecessarily dirty metadata.
- [ ] Same stable Catch ID is retained.
- [ ] Owner/provenance/location privacy boundaries remain intact.
- [ ] Offline edits/add/remove survive close/reopen.
- [ ] Pending changes sync after reconnect.
- [ ] Location/privacy is manageable through existing privacy UI.
- [ ] Healthy sync remains visually quiet.
- [ ] Unexpected caught exceptions are logged safely.
- [ ] EN/FR localisation is complete.
- [ ] Required automated tests are green.
- [ ] Web build/publish succeeds.
- [ ] Full `.claude/rules/self-review.md` review is completed.
- [ ] No unrelated scope is added.

kingkeamo:
Closes #20.

## Summary

- Rebuilt `/catches/{id}/edit` around `MudExpansionPanels` (Photographs / Catch details / Location & privacy) as the single mobile-first Catch Details/Edit experience the Logbook opens into.
- Extracted a shared `Features/Catch/Components/CatchPhotographCarousel` component (index/nav/wrap-around/"N of N"/placeholder/local-bytes-vs-`RemoteUrl`/alt text, optional `Editable` remove control) with no persistence/sync knowledge of its own, and migrated `CatchCard` onto it (ids preserved exactly, all 32 existing tests unchanged).
- Photographs can now be added and individually removed from an existing Catch, fully offline-first through the existing `ICatchStore`/`ICatchSynchroniser` — no parallel persistence path. Removing the last visible photo is blocked in the UI; every other removal goes through a destructive confirm dialog.
- New server-side delete slice: `DELETE /api/catches/{catchId}/photographs/{photographId}` → `DeleteCatchPhotographCommand` → `CatchPhotographService.DeleteAsync` → `ICatchRepository.DeletePhotographAsync`, ownership-scoped via `ICurrentUser`, idempotent. R2 object is deleted *before* the DB row (deliberate — there's no reconciliation job for orphaned R2 objects in this codebase, so this ordering guarantees the two stores never silently diverge; failures at either step are safely retryable since S3 delete-of-missing-key is a no-op).
- New `SyncStatus.PendingDeletion`; `CatchSynchroniser` branches on it — success splices the photo out locally and calls the new `ICatchClient.DeletePhotographAsync` (404 = idempotent success); failure keeps it `PendingDeletion` so retries keep routing through delete, never re-upload.
- `CatchEdit` now falls back to the server when a catch isn't in the local IndexedDB (e.g. a "server-only" catch merged into the Logbook by #100 but never localised): downloads the catch and every photo's bytes, adopts it locally as `Synchronised`, then behaves exactly like any other local catch from that point on — reusing the existing store/sync architecture rather than special-casing anything downstream.
- Location & privacy section reuses the existing `LocationPrivacyModal`/`IModalService` unchanged.

## Self review

```
Self review:
- Issue/AC: PASS — all Definition of Done items satisfied; two items deliberately deferred (see below), both explicitly anticipated/permitted by the issue text itself
- Architecture/CQRS: PASS — Endpoint → IMediator → Handler → Service → Repository chain intact for the new delete slice; Web never references Application/Infrastructure
- Rules: PASS
- Security/privacy: PASS — delete is ownership-scoped via trusted ICurrentUser, never client-supplied; R2 deletion never exposes credentials/keys to Web; no photo bytes/coordinates/tokens logged
- Database/migrations: N/A — no schema change; CatchPhotograph table already existed, only a new DELETE query against it
- Tests/coverage: PASS — 1113 .NET tests + 138 JS tests, all green
- Browser: N/A (reason: bUnit covers component behaviour with mocked clients; CatchEdit uses production Cognito auth, and testing-blazor.md explicitly disallows test-only auth/dual-host hacks to force Playwright onto a product ticket; existing Playwright suite targets service-worker/PWA-shell behaviour, not per-feature UI)
- Web/UI/localisation: PASS — EN/FR added for every new string; existing keys reused where semantically identical (photo alt text, take/choose photo, no-location, visibility labels)
- Code smells: PASS after fixes (see below)
- CI/deployment: PASS — no migration/Terraform/CI changes needed; `dotnet publish` verified green

1. `MudExpansionPanels Gutters="false"` on CatchEdit stripped all panel padding (MudBlazor's own docs: Gutters "adds left and right padding", defaults true) — every field/chip/photo sat flush against the panel edge.
2. The carousel's "large" variant had no size cap, rendering a full-width 1:1 square photo — visually oversized on the Edit page.
3. CatchEdit had no path for a catch that's visible in the Logbook (server-only, merged in-memory by #100) but never saved to this device — editing it always failed to load.
4. A real bug in the existing offline JS layer: `putCatchWithPhotographs` never deleted a photo blob that dropped out of a later write, and the read-side merge (`orderPhotographs`, mirrored in `CatchJson.OrderPhotographs`) then resurrected the orphaned blob on the next read — a locally "deleted" photo would silently reappear after reload.
5. `DeleteCatchPhotographCommand` was missing the `IsObjectStorageConfigured` guard present on every other photograph-storage command.
6. `CatchRepository`'s multi-photograph LEFT JOIN grouping (`GetByUserIdAsync`/`GetByIdAsync`) had no test exercising more than one photo per catch — the exact case the JOIN's multi-mapping exists for.

Fixes made:
1. Removed `Gutters="false"`.
2. Capped the carousel's large variant to a fixed 14rem height / 20rem max-width, centred (mirrors Record Catch's existing preview treatment).
3. Added `ICatchClient.GetAsync`/`DownloadPhotographAsync` and `CatchEdit.TryLocalizeFromServerAsync`: fetches the owned catch + downloads every photo, then `CatchStore.SaveAsync` adopts it locally before editing continues. A distinct "not saved on this device yet, connect and try again" message covers the case where localisation itself can't complete (still offline, expired URL).
4. Fixed `putCatchWithPhotographs` to delete photo-store blobs whose id drops out of the incoming list, with a regression test; switched `CatchSynchroniser`'s delete-completion path from `UpdateSyncStateAsync` (can only ever refresh existing metadata entries, never shrink the list) to `SaveAsync` (full replace).
5. Added the missing guard + test.
6. Added Testcontainers coverage for 3-photo grouping and two-catch cross-contamination; caught and fixed a GUID collision of my own between the two new tests (identical hardcoded photograph ids colliding on `CatchPhotograph`'s primary key against the shared Postgres instance) — switched to `Guid.NewGuid()`.

Remaining deliberate gaps:
- Record Catch was **not** migrated onto the shared carousel component. Its swipe-gesture, clamp-boundary (non-wrapping) navigation and pre-save `PendingPhotograph` model (no `CatchId` yet) are meaningfully different from the shared component's contract; forcing it in meant either changing its UX or growing the component's parameter surface — both against the addendum's own "avoid a generic configuration object" guidance.
- No background reconciliation job for an R2 object that fails to delete after a successful DB delete — the issue's own text permits this ("if no suitable retry mechanism exists, stop... and document the smallest safe choice"); none exists in this codebase, so the failure is logged (CatchId/PhotographId only) rather than silently orphaned with no record, and the delete-before-DB ordering keeps the two stores from ever diverging on a partial failure.
- No custom focus management after inline photo add/remove beyond MudBlazor's own defaults.
- No dirty-navigation warning before leaving the page with unsaved edits — the issue marks this optional and explicitly discourages new navigation-state infrastructure for this ticket.
- `CatchList`'s existing server-only-catch mapping still hardcodes `Location: null` — pre-existing gap from #100, unrelated to this ticket, not fixed here.
```

## Test plan

- [x] `dotnet build FishingLogBook.sln`
- [x] `dotnet publish src/FishingLogBook.Web` (Release)
- [x] `dotnet test FishingLogBook.sln` — 1113 passed, 0 failed
- [x] `npm test` (vitest, `src/FishingLogBook.Web`) — 138 passed, 0 failed
- [x] `dotnet format FishingLogBook.sln` — clean


kingkeamo:
Review against #20: the overall shape is strong and CI is green, but I see two blockers before final dev/device testing.

1. **The server-side R2/DB delete ordering contradicts #20 and the implementation plan posted on the issue.** `CatchPhotographService.DeleteAsync` currently deletes the R2 object first and only then deletes the `CatchPhotograph` row; the new test explicitly locks that ordering in. #20 says that if DB removal succeeds but R2 cleanup fails, the photograph must stay removed from the Catch, the failure must be logged safely, and cleanup should be retried/handled using the smallest existing mechanism (or the limitation documented if none exists). The implementation plan likewise proposed DB deletion as authoritative with best-effort R2 cleanup. With the current R2-first ordering, a transient DB failure after a successful object delete leaves a persisted photograph row pointing at a missing object until a later retry. Please bring the implementation back into line with the ticket/plan, or update #20 with an explicit agreed design change before retaining the reverse ordering. Add coverage for the partial-failure case that proves the chosen semantics.

2. **New `CatchEdit` exception paths silently swallow unexpected failures.** `TryLocalizeFromServerAsync`, photo add/save, photo remove/save and related new broad `catch (Exception)` blocks only flip UI flags/return null; they do not log the actual exception. #20 explicitly requires every touched/new try/catch to log unexpected exceptions before showing generic safe UI, while avoiding photo bytes, precise coordinates, presigned URLs, tokens, etc. Please route these through the existing safe logging/diagnostics boundary (and preserve caller-cancellation handling rather than reporting normal disposal as an error). Also include coverage that failures produce the safe UI state without leaking sensitive data.

Everything else I checked is directionally aligned: ownership is resolved server-side, the delete route does not trust a client object key, the tombstone/pending-deletion flow is local-first and retryable, CatchCard now uses the shared carousel, server-only catches are localised before editing, and the IndexedDB orphan-photo resurrection bug has regression coverage. Once the two points above are addressed, I’d do the final phone/desktop interaction pass called for by #20 (carousel/expansion panels/no horizontal overflow) before merging.

kingkeamo (commented):
Latest sweep looks good overall. The CatchEdit split into focused partials is a real readability improvement, the shared photograph component is now also used by RecordCatch, the server-localisation exception paths are logged, the IndexedDB stale-photo cleanup remains covered, and the Codex setup files do not contain credentials (they reference an environment variable only). CI is green.

One blocker remains, and it looks like the cleanup sweep has actually regressed the photograph-delete ordering we agreed earlier:

`CatchPhotographService.DeleteAsync` currently does **DB delete first**, then R2 delete, and if R2 deletion throws it logs the exception and still returns success. The new tests explicitly assert that behaviour (`ItShouldKeepTheDatabaseDeletionWhenObjectStorageCleanupFails` and `operations.Should().Equal("database", "object-storage")`). That means the client tombstone is cleared even though the private R2 object can remain orphaned permanently — exactly the failure mode we wanted to avoid because there is no reconciliation job.

Please restore the intended retry-safe order:

1. Load/verify owned photograph.
2. Build the trusted object key server-side.
3. Delete the R2 object first.
4. If R2 deletion fails, log + return failure so the local `PendingDeletion` tombstone remains and retry can occur.
5. Delete the DB row only after R2 deletion succeeds.

S3/R2 delete is idempotent, so if object deletion succeeds and the DB delete then fails, the existing client tombstone can safely retry the same operation; deleting an already-missing object remains a no-op. Update the service tests to assert `object-storage` then `database`, and that an R2 failure does **not** call the repository delete and returns failure.

On the previous rows-affected point: after re-reading the current service flow, I no longer think that needs to block this PR. `LoadOwnedPhotographAsync` establishes ownership/existence first and the public delete operation is deliberately idempotent, so a zero-row delete caused by a race/retry can reasonably remain success.

Once the delete ordering is restored, I don't currently see another blocker.

Issue #132, opened 2026-08-22
PWA | Investigate and design secure offline cold-start access


kingkeamo:
## Context

CBDF's offline catch workflow is now strong once the installed PWA is already running: catches can be listed, recorded and edited locally, photographs/local state are preserved, and queued work can sync when connectivity returns.

A remaining gap has been identified on real devices:

**If the installed PWA is fully closed/killed and then opened with no network connection, the user cannot reliably get back into the authenticated app.**

This is not primarily an offline-data problem. The published service worker is intended to cache and serve the application shell, and catch data already lives locally. The missing architectural piece appears to be **offline restoration/unlock of the previously authenticated local user** during a cold start.

Current authentication uses the standard Blazor WebAssembly OIDC stack. No custom offline authentication/session restoration mechanism has been identified yet.

This matters directly to the product: an angler may leave home with connectivity, travel somewhere remote, have the OS kill/suspend the PWA, then need to reopen CBDF with no signal. They must still be able to access the locally supported fishing workflow.

## Goal

Investigate and design the **smallest secure approach** that allows a previously authenticated user to cold-start the installed CBDF PWA with no network and access only that user's locally supported functionality/data.

This is an **architecture spike first**. Do not jump straight to implementing a custom offline-auth system.

The preferred end-state is conceptually:

1. User authenticates normally online through the existing OIDC/Cognito flow.
2. CBDF establishes a secure local offline-access entitlement for that authenticated user/device.
3. Later, the PWA is completely closed.
4. Device has no network connection.
5. User opens CBDF from the installed icon.
6. Cached app shell starts.
7. CBDF securely restores/unlocks the previously authorised local user.
8. Offline mode exposes only locally supported functionality:
   - Catch list
   - Record catch
   - Edit catch
   - locally available catch photographs/data
9. API-dependent functionality remains unavailable/disabled while offline.
10. When connectivity returns, normal OIDC/server authentication is restored/revalidated and queued work synchronises.

## Device authentication feasibility — investigate first

Investigate whether the ideal offline unlock can use **device authentication** from an installed PWA, including:

- Face ID / Touch ID on iPhone/iPad;
- fingerprint / face / device PIN on Android;
- WebAuthn/passkeys/platform authenticators;
- any other standards-based browser mechanism that can safely prove local user presence without a network round trip.

The desired UX, if technically reliable, is approximately:

**Open CBDF offline → Unlock → Face ID / fingerprint / device PIN → Catches / Record Catch**

[...]

PR #139, opened 2026-08-23
FLB#132 | Keep offline capability probe available without authentication

kingkeamo:
## Summary

- routes the disposable WebAuthn capability probe without waiting for remote authentication state
- allows the Landing probe action to work after an offline iPhone PWA cold start
- keeps all existing `[Authorize]` application routes unchanged
- adds a router-level regression test with authentication deliberately left pending

## Validation

- focused Landing/probe tests: 20 passed
- complete Web suite: 680 passed
- Release solution build: 0 warnings / 0 errors
- `git diff --check`: passed

Follow-up to #132. This contains the physical-device finding discovered after PR #138 had already been merged.

### DIFF
--- a/src/FishingLogBook.Web/App.razor
+++ b/src/FishingLogBook.Web/App.razor
@@ -5,7 +5,8 @@
     <CascadingAuthenticationState>
         <Router AppAssembly="@typeof(App).Assembly" NotFoundPage="typeof(FishingLogBook.Web.Pages.NotFound.NotFound)">
             <Found Context="routeData">
-                @if (routeData.PageType == typeof(FishingLogBook.Web.Features.Onboarding.Pages.Landing.Landing))
+                @if (routeData.PageType == typeof(FishingLogBook.Web.Features.Onboarding.Pages.Landing.Landing)
+                    || routeData.PageType == typeof(FishingLogBook.Web.Features.Diagnostics.Pages.WebAuthnCapabilityProbe.WebAuthnCapabilityProbe))
                 {
                     <RouteView RouteData="@routeData" DefaultLayout="@typeof(MainLayout)" />
                 }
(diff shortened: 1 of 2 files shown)


kingkeamo:
Probe-only PRF decryption investigation update:

Diagnosis:
- Provisioning stored the credential ID and PRF salt, but did **not** create the AES-GCM harmless-payload envelope.
- That envelope was only created by the separate **Verify online credential** action.
- The probe nevertheless enabled offline testing as soon as provisioning metadata existed, so an offline GET could correctly return a PRF result while payload verification collapsed a missing envelope into the generic “Harmless payload decrypted: No”.
- `navigator.onLine` is only a browser connectivity hint on iOS and is not proof of internet reachability, which explains “Offline at invocation: No” with radios disabled.

Probe-only fix:
- Offline unlock remains disabled until online verification has created a complete envelope.
- The UI now distinguishes missing envelope, PRF mismatch, invalid persisted envelope, AES-GCM decrypt failure, plaintext mismatch, and successful verification.
- Safe diagnostics report PRF result representation, byte length, and equality of a truncated SHA-256 fingerprint. Actual PRF material is never logged or persisted.
- ArrayBuffer and typed-array PRF results are both handled.
- Network wording now explicitly labels `navigator.onLine` as a non-authoritative browser hint.
- No Cognito, AuthenticationStateProvider, token storage, offline entitlement, service worker, catch storage, or sync code changed.

Validation:
- JavaScript: 226 passed
- Web tests: 681 passed
- Release build: succeeded, 0 warnings / 0 errors
- dotnet format and git diff checks: clean

Required real-device retest after deployment:
online provision → **verify online** → kill installed PWA → Wi-Fi/cellular off → reopen → explicit offline unlock. Expected: GET/UV/PRF succeed and harmless payload decrypts Yes. If it still fails, the new diagnostic outcome will identify whether CREATE/GET PRF-derived material differs or whether envelope decoding/decryption failed.

Issue #172, opened 2026-08-26
Trips T1 | Trip domain, persistence and API foundation


kingkeamo:
Architecture/product baseline: **#108 — Agreed V1 implementation baseline**. Do not repeat repository-wide investigation unless current code contradicts that baseline.

Part of #108. First slice: the durable server/domain foundation, with no UI and no offline surface, so it can be reviewed purely as a domain and schema decision.

## Goal

A Trip can be created, updated and read through the API, owned by one angler, with optional title and optional first-class location, enforcing one active trip per owner.

## Scope

- `Domain/Trips/`: `Trip`, `TripLocation`, `TripStatusEnum`.
- `Shared/Dtos/`: trip DTOs following existing `CatchDto` / `CatchViewDto` conventions.
- Migration creating `"Trip"` with the partial unique active index and an `EndedOn >= StartedOn` check.
- `UpsertTripCommand` + validator, `GetTripQuery`, `GetMyTripsQuery`, `ITripService`, `ITripRepository`, Mapster registration.
- `POST /api/trips`, `GET /api/trips`, `GET /api/trips/{tripId:guid}` using existing endpoint/error-to-status conventions.
- `docs/Requirements.md` section for solo trips — none exists today (only §25–27 guided).

## Location shape

Location is V1. Three concepts must stay distinct and unconflated:

- **exact coordinates** — optional, private by default, with capture provenance;
- **display place** — optional human-readable place name;
- **Venue/Fishery** — future, not implemented.

The schema must not need replacing when searchable place selection arrives. No Google Maps, no place lookup, no reverse geocoding, no mapping, no Venue table.

## Out of scope

Trip UI · IndexedDB/local store (T2) · synchroniser (T4) · active-trip banner (T3) · `Catch.TripId` implementation (T5 — keep relationship work there unless the foundational migration genuinely requires the column now) · trip photographs (T6) · notes (T7) · timeline (T8) · Plan Trip UI · place search · reverse geocoding · participants · video.

## Acceptance criteria

- [ ] Given no catches, photographs or notes, when a Trip is created, then it persists and is valid.
- [ ] Given no title, when a Trip is created, then it persists with a null title.
- [ ] Given no location, when a Trip is created, then it persists with no location.
- [ ] Given a valid location, when a Trip is created, then coordinates, accuracy, capture time, source, visibility and consent version persist and read back unchanged.
- [ ] Given a place name and no coordinates, when a Trip is created, then it persists.
[...]

Issue #166, opened 2026-08-26
Offline storage | Stop IndexedDB photo reads freezing online catch pages


kingkeamo:
## Problem

Physical testing shows the app can freeze / become effectively unusable after login when loading the Catches page and when editing a catch.

The captured console log shows the backend/auth path is responsive, while local IndexedDB catch reads/writes are timing out. This should be treated as a stability blocker rather than folded into the presentation-only cleanup in #164.

## Observed behaviour

Typical sequence after login:

- authentication succeeds;
- profile/catalogue/API calls return successfully and quickly;
- local catch logbook read times out;
- catch list reports it cannot load;
- pending catch synchronisation starts;
- a photograph upload-url request for a stale/pending catch returns 404 and is retried;
- sync-state IndexedDB write then times out;
- normal server catch GETs and photograph/R2 downloads still succeed.

Editing an existing catch can trigger the same local-store path and the application appears frozen while it waits.

This strongly suggests the browser-side local storage/photo loading path is the bottleneck rather than the API.

## Likely root cause

The current local catch-store API is too coarse-grained.

`CatchStore.GetAsync(ownerUserId, catchId)` obtains a single catch by calling `GetAllAsync(...)` and then selecting the requested catch.

The JavaScript `getAllCatchesWithPhotographs(ownerUserId)` implementation:

1. reads every catch from IndexedDB;
2. scans the photograph store;
3. loads photograph binary data for all visible catches;
4. converts each photograph byte array to Base64 in JavaScript;
5. sends those Base64 strings across JS/WASM interop;
6. .NET reconstructs the photograph byte arrays.

This means operations that only need catch metadata — including opening one catch — can repeatedly load and encode every locally stored photograph.

The same expensive read can also occur from several paths around login/reconnect/synchronisation and `StateChanged` refreshes.

## Timeout concern

Review timeout ownership as part of the fix.

The .NET local-store operation can time out before the JavaScript IndexedDB transaction timeout expires. That can leave the underlying JS operation continuing after the caller has already abandoned it, while subsequent reads/writes begin.

Do not solve this merely by increasing the timeout. Normal metadata reads should be comfortably fast.

## Required changes

Keep the proven offline architecture, but make local storage access appropriately granular.

- Add a true single-catch IndexedDB read for `GetAsync`; do not implement it via `GetAllAsync`.
- Separate catch metadata/list reads from photograph-byte reads.
[...]

Issue #108, opened 2026-08-21
Fishing Trips | Capture the whole day, not just the catch


kingkeamo:
## Product idea

Introduce **Fishing Trip** as a first-class concept in Catch, But Don't Forget.

A Catch records a fish. A **Trip records the experience of going fishing** — including the many days where no fish are caught at all.

This should become the next major domain feature after the current onboarding/authentication work is stable and properly validated. Do not start implementation until the onboarding work is tidied up and working reliably on the target devices.

The core product idea is:

> **The Trip is the fishing experience. The Catch is what happened during it.**

This gives the product a natural spine for future venues/fisheries, guides, clubs, competitions and bookings without making those separate disconnected modes.

## Why this matters

Anglers do not remember their fishing lives purely as individual catch records.

They remember:

- that early morning on Corrib;
- the unreal sunrise before anyone caught anything;
- the completely blank day in terrible weather;
- the day out with friends;
- the huge trout someone lost beside the boat;
- the photos, videos, jokes and stories around the fishing;
- the one exceptional fish caught during an otherwise ordinary day.

A trip with **zero catches** is still a meaningful fishing record and must be treated as such.

A few years later, someone should be able to say:

> "Remember that day we had on Corrib?"

and the angler can open Catch, But Don't Forget and find the trip: who was there, when and where they went, the sunrise photo, catches, videos, trip photos and notes from the day.

This extends the meaning of the brand naturally:

**Catch · Release · Remember**

Not only remembering the fish — remembering the days spent fishing.

## Core domain principle

Fishing Trip must be a **first-class domain record**, not merely a container around `Catch[]`.

At minimum a trip needs to be capable of owning/associating:

- owner/creator;
- participants;
- planned/start/end times;
- optional name/title;
- optional place/location context;
- catches;
- trip photos;
- trip videos;
- trip notes;
- trip status/lifecycle;
- privacy/sharing choices.

The detailed model should be designed before implementation rather than assuming the above maps one-to-one to database tables.

### 1. Start fishing now

The fastest possible journey.

An angler should be able to tap something equivalent to:

**Start fishing**

and begin a trip with minimal/no form filling.

The app can capture the start time automatically and location according to the user's existing location/privacy choices.

The angler should immediately reach an **Active Trip** experience.

### 2. Plan a trip

An angler can create a future trip, for example:

**Saturday morning on Corrib**

Useful initial information might include:

- date/time;
- optional location/place;
- optional title;
- friends to invite.

[...]

PR #182, opened 2026-08-27
FLB#173 | Local Trip store and IndexedDB v5

### LINKED ISSUE #173: Trips T2 | Local Trip store and IndexedDB v5

Part of #108. Depends on #172 (T1) for the model shape only; can be developed in parallel.

## Goal

Trips persist locally in IndexedDB with owner partitioning, following the granular read pattern established by #166, without disturbing existing catch data.

## Scope

- IndexedDB **v4 → v5**, additive only: `trips` store (`keyPath: 'id'`).
- `ITripStore` / `IndexedDbTripStore` in `Features/Trips/Offline/Stores/`.
- JS module functions beside the existing catch store functions.
- Metadata-only read for lists; single-trip read by key; owner filtered **during** the cursor, not after.
- Local `TripModel` carrying `SyncStatus` and `SyncedAt`, matching catch conventions.

## Critical constraint

The upgrade touches a database already holding real queued catches and photographs on the test devices. It must be **purely additive** — `createObjectStore` guarded by `objectStoreNames.contains`, no store deleted, no record rewritten. **No user's queued offline catch may be lost.**

## Out of scope

Synchroniser (T4) · any UI (T3) · trip photograph blobs (T6) · notes (T7) · cleanup wiring beyond the store method itself.

## Acceptance criteria

- [ ] Given an existing v4 database with catches and photograph blobs, when the app opens after this change, then the database is at v5 and every existing catch and blob is intact.
- [ ] Given a Trip saved locally, when read back by id, then all fields including location round-trip.
- [ ] Given trips belonging to two owners, when read for one owner, then only that owner's trips are returned.
- [ ] Given an empty owner id, when a read is attempted, then it is rejected.
- [ ] Given several stored trips, when the metadata list is read, then no photograph blob store is opened.
- [ ] Given a trip that is not stored, when read by id, then null is returned.
- [ ] Given an active trip, when the owner's active trip is queried, then it is found from stored data rather than component state.

## Test requirements

- vitest against `fake-indexeddb` for the JS module, including owner isolation and the not-found path.
- .NET store tests following the existing `CatchStoreTests` pattern.
- **Browser (Playwright): seed a v4 database with catches and photograph blobs, upgrade to v5, assert nothing was lost.** This is the single most important test in the slice.
- Instrument `IDBObjectStore` and prove the trip metadata path reads no blobs, carrying forward the #166 discipline.

## Definition of done

kingkeamo:
Closes #173
Part of #108

Second Trips slice, on top of #172 (now merged). Architecture/product baseline: [#108 — Agreed V1 implementation baseline](https://github.com/kingkeamo/fishing-logbook/issues/108#issuecomment-5428632438).

Trips now persist locally, so an angler can start, continue and finish one without connectivity, and an active trip survives closing and reopening the app. No UI, no synchroniser, no `Catch.TripId` — those are T3, T4 and T5.

## The IndexedDB upgrade

**Version 4 → 5 on the `FishingLogBook` database.** This is the highest-risk change in the whole feature: it touches a database already holding real queued catches and photograph blobs on the test devices.

Purely additive. One `createObjectStore` guarded by an existence check. Nothing renamed, dropped, cleared or rewritten, and the pre-existing legacy `testCatches` cleanup is carried over verbatim rather than reworked.

The other three databases — `FishingLogBookOfflineAccess`, `FishingLogBookPreferences`, `FishingLogBookDiagnostics` — are separate and untouched, which a test asserts rather than assumes.

### One store only

`trips` (`keyPath: 'id'`, no indexes). **`tripPhotographs` and `tripNotes` are deliberately not created.**

The upgrade handler is a series of `if (!contains) create` guards, so T6 and T7 can each add their store in a later additive version, and a v4 user who lands on v7 gets all of them in a single upgrade. There is no safety benefit to creating empty stores now, so doing it would be pure speculation.

### Shared schema ownership

The database name, version, upgrade handler and transaction helpers move out of `catch-store.js` into a new `logbook-database.js` that both stores import. Trips and catches genuinely share one database, and creating a `trips` store inside a function called `openCatchDatabase`, governed by a constant called `CATCH_DATABASE_VERSION`, would have been actively misleading.

`catch-store.js` re-exports its previous names, so **no Catch behaviour changed** — all 41 existing Catch vitest tests pass untouched.

## Local model and store

```csharp
TripModel(Id, OwnerUserId, Status, StartedOn, EndedOn?, Title?, PlaceName?,
          Location?, SyncStatus, SyncedAt?)

ITripStore
  SaveAsync(trip, ct)
  GetAllAsync(ownerUserId, ct)
  GetAsync(ownerUserId, tripId, ct)      // direct key read
  GetActiveAsync(ownerUserId, ct)        // direct owner + status read
```

`Status` is a `string` constrained by `TripConstants.IsKnownStatus`, mirroring how `CatchLocationModel.Visibility` already works — Web references Shared only, so the Domain enum is unavailable and duplicating it would be worse than a constrained string.

**No `CreatedOn`/`UpdatedOn` locally.** They are server-assigned and nothing in T3 or T4 needs them, so adding them now would be inventing fields for later slices.

**Sync state is two fields**: `SyncStatus` so T4 can find pending work, and `SyncedAt` because #108's retention rules key on it. No `MetadataSyncStatus` — trips own no child media in this slice.

### No read is implemented by listing and filtering

`GetAsync` is a direct `store.get(tripId)`; `GetActiveAsync` is a single owner-scoped cursor that stops at the first match. This is the #166 anti-pattern and it is not repeated — a browser test instruments `IDBObjectStore` and asserts a single-trip read performs exactly one `get` against `trips` and nothing else.

## One active trip locally

Enforced **inside the readwrite transaction that writes the trip** — one authoritative local write boundary, not split across layers and not dependent on any component state.

The JS returns a discriminated outcome (`'saved'` / `'activeConflict'`) which the store turns into `TripAlreadyActiveException`. A returned outcome rather than error-message matching, so an expected domain conflict is distinguishable from a genuine failure without parsing strings.

| Scenario | Behaviour |
|---|---|
| Second **distinct** active trip, same owner | rejected, nothing written |
| Re-saving the **same** active trip | succeeds |
| New active trip after the earlier one completed | succeeds |
| Each owner holding their own active trip | succeeds |
| Write moving an existing trip id to another owner | rejected outright |

## Owner isolation

`ownerUserId` is filtered **during** every cursor in JS — never after an expensive read — with a `LocalTripVisibility.ForOwner` pass in .NET as defence in depth. Proven for all three reads plus the ownership-reassignment case.

## Cold reopen

No in-memory current-trip state exists anywhere. The active trip is discovered from owner-scoped stored data, proven by a real `page.reload()` followed by `getActiveTrip`. This is what makes T3's Start Fishing survive a force-quit on the water.

## Cleanup

Verified rather than assumed: `cleanupSyncedCatches` opens a transaction scoped to `[catches, catchPhotographs]` and cursors only the catch store, so it **cannot** reach `trips`. No guard was needed. Actual trip retention belongs to T4, where the #108 rules apply — active and pending trips never evicted.

## Tests

**26 browser tests** (13 specs × Chromium and WebKit) covering all 18 assertions #173 asked for: the v4→v5 upgrade with catch metadata and a 64 KB blob surviving, the entitlement database untouched, blank / active / completed / with-location / without-location round-trips, location provenance preserved, direct-read instrumentation, owner scoping on all three reads, the four active-trip variants, and reload recovery.

**33 .NET tests**: 10 `TripJson` round-trip, 12 store and JS-interaction (including the timeout path), 2 DI resolution, plus 9 updated schema guards.

Full validation: `dotnet format` clean · `dotnet build` clean · **1,748 .NET tests** · vitest **280** · browser **90** · `git diff --check` clean.

No credentialed E2E — this slice touches no user journey. No physical-device acceptance claimed; that is T9.

## Self review

```text
Self review:
- Issue/AC: PASS — every acceptance criterion implemented and tested
- Architecture: PASS — feature-first under Features/Trips, store role in Offline/Stores
- Rules: findings 1, 2 (fixed)
- Security/privacy: PASS — owner scoped during every cursor; location provenance and
  visibility round-trip intact; no coordinates or payloads logged
- Database/migrations: PASS — additive IndexedDB upgrade, no server migration in this slice
- Tests/coverage: PASS — real browser proof, not mocks, for the upgrade
- Browser: PASS — 26 tests across two engines
- Web/UI/localisation: N/A — no user-visible text in this slice
- Code smells: findings 3, 4 (fixed)
- CI/deployment: PASS — no manual steps
```

**Findings discovered during self-review:**

1. **A deliberate tripwire fired, exactly as designed.** `schema-safety.test.js` asserted `CATCH_DATABASE_VERSION === 4` and an exact store set — a guard against casual schema bumps, which #108 explicitly warned about.
2. **Two .NET tests read the JS source** (`JavaScript/OfflineStore`) and broke, because extracting the shared module moved the timeout and diagnostics code they inspect.
3. **Enum casing mismatch, caught before it shipped.** I first used `JsonStringEnumConverter()` with no naming policy, producing `"SavedLocally"`, while the existing JS reads `'synchronised'`. Left alone this would have recreated exactly the class of latent bug found in #166, where JS and .NET disagreed on a persisted enum's shape and the tests did not reflect production.
4. **`ITripStore` had no consumer**, so nothing proved it was registered — a typo would only have surfaced in T3.

**Fixes made:**

1. Updated the guards **deliberately** to v5 and the three-store set, kept as exact-set assertions so a future casual change still trips them, and **added two more**: trip-store must not mention other features' schema names, and the database name must live in exactly one module.
2. Added `logbook-database.js` and `trip-store.js` to the concatenated source, so those assertions now cover the whole offline-storage surface. Strengthened, not weakened.
3. Aligned to `JsonNamingPolicy.CamelCase` and added a test asserting the serialised JSON contains `"syncStatus":"synchronised"` — locking the exact cross-language contract.
4. Added DI resolution tests, which immediately hit the `IAsyncDisposable` container-disposal trap and were fixed with `CreateAsyncScope`.

**Remaining deliberate gaps:**

- Trip retention is not wired into cleanup. Verified impossible for the current cleanup to touch trips, so this is a genuine T4 concern rather than a hole.
- No UI, synchroniser, `Catch.TripId`, photographs, notes or timeline — all later slices, all untouched here.

## Manual post-merge steps

**None.** No database migration, no Terraform, no configuration. The IndexedDB upgrade runs in each browser on next open.

### DIFF
--- a/src/FishingLogBook.Web/BrowserTests/logbook-upgrade.spec.js
+++ b/src/FishingLogBook.Web/BrowserTests/logbook-upgrade.spec.js
@@ -1,0 +1,368 @@
+import { expect, test } from '@playwright/test';
+
+const tripStoreModule = '/src/FishingLogBook.Web/wwwroot/js/storage/trip-store.js';
+const catchStoreModule = '/src/FishingLogBook.Web/wwwroot/js/storage/catch-store.js';
+const logbookModule = '/src/FishingLogBook.Web/wwwroot/js/storage/logbook-database.js';
+const databaseName = 'FishingLogBook';
+const entitlementDatabaseName = 'FishingLogBookOfflineAccess';
+const ownerUserId = '11111111-1111-1111-1111-111111111111';
+const otherUserId = '22222222-2222-2222-2222-222222222222';
+const seededCatchId = 'cccccccc-cccc-cccc-cccc-cccccccccccc';
+const seededPhotographId = 'dddddddd-dddd-dddd-dddd-dddddddddddd';
+const tripId = 'aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa';
+const otherTripId = 'bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb';
+const startedOn = '2026-08-26T05:32:00+00:00';
+
+test.beforeEach(async ({ page }) => {
+    await page.goto('/src/FishingLogBook.Web/BrowserTests/harness/index.html');
+    await page.evaluate(async (names) => {
+        for (const name of names) {
+            await new Promise((resolve) => {
+                const request = indexedDB.deleteDatabase(name);
+                request.onsuccess = () => resolve();
+                request.onerror = () => resolve();
+                request.onblocked = () => resolve();
+            });
+        }
+    }, [databaseName, entitlementDatabaseName]);
+});
+
+async function seedVersionFour(page) {
+    return page.evaluate(async ({ databaseName, seededCatchId, seededPhotographId, ownerUserId }) => {
+        const db = await new Promise((resolve, reject) => {
+            const request = indexedDB.open(databaseName, 4);
+            request.onupgradeneeded = () => {
+                const upgrading = request.result;
+                upgrading.createObjectStore('catches', { keyPath: 'id' });
+                upgrading.createObjectStore('catchPhotographs', { keyPath: 'id' });
+            };
+            request.onsuccess = () => resolve(request.result);
+            request.onerror = () => reject(request.error);
+        });
+
+        await new Promise((resolve, reject) => {
+            const transaction = db.transaction(['catches', 'catchPhotographs'], 'readwrite');
+            transaction.objectStore('catches').put({
+                id: seededCatchId,
+                userId: ownerUserId,
+                caughtOn: '2026-06-14T09:48:00+00:00',
+                notes: 'the one from before the upgrade',
+                syncStatus: 'savedLocally',
+                photographs: [{ id: seededPhotographId, catchId: seededCatchId, contentType: 'image/jpeg' }]
+            });
+            transaction.objectStore('catchPhotographs').put({
+                id: seededPhotographId,
+                catchId: seededCatchId,
+                contentType: 'image/jpeg',
+                bytes: new Uint8Array(64 * 1024).fill(7).buffer
+            });
+            transaction.oncomplete = () => resolve();
+            transaction.onerror = () => reject(transaction.error);
+            transaction.onabort = () => reject(transaction.error);
+        });
+
+        const version = db.version;
+        const stores = [...db.objectStoreNames].sort();
+        db.close();
+        return { version, stores };
+    }, { databaseName, seededCatchId, seededPhotographId, ownerUserId });
+}
+
+async function seedEntitlementDatabase(page) {
+    return page.evaluate(async (name) => {
+        const db = await new Promise((resolve, reject) => {
+            const request = indexedDB.open(name, 1);
+            request.onupgradeneeded = () => {
+                request.result.createObjectStore('deviceEntitlements', { keyPath: 'ownerKey' });
+            };
+            request.onsuccess = () => resolve(request.result);
+            request.onerror = () => reject(request.error);
+        });
+
+        await new Promise((resolve, reject) => {
+            const transaction = db.transaction('deviceEntitlements', 'readwrite');
+            transaction.objectStore('deviceEntitlements').put({
+                ownerKey: 'owner-1',
+                state: 'ready',
+                ciphertext: [1, 2, 3, 4]
+            });
+            transaction.oncomplete = () => resolve();
+            transaction.onerror = () => reject(transaction.error);
+        });
+
+        const version = db.version;
+        db.close();
+        return version;
+    }, entitlementDatabaseName);
+}
+
+function newTrip(overrides = {}) {
+    return {
+        id: tripId,
+        ownerUserId,
+        status: 'Active',
+        startedOn,
+        endedOn: null,
+        title: null,
+        placeName: null,
+        location: null,
+        syncStatus: 'savedLocally',
+        syncedAt: null,
+        ...overrides
+    };
+}
+
+async function saveTrip(page, record) {
+    return page.evaluate(async ({ tripStoreModule, record }) => {
+        const { putTrip } = await import(tripStoreModule);
+        return putTrip(JSON.stringify(record));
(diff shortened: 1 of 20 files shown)


kingkeamo (commented):
Reviewed T2 against #173 and the #108 implementation baseline, including the actual changed-file inventory and the storage/upgrade code paths.

No blocker found.

Key points checked:
- IndexedDB upgrade is additive v4 -> v5 and only adds `trips`; existing Catch/photo stores are preserved.
- shared `logbook-database.js` keeps schema/version ownership in one place while `catch-store.js` preserves its existing public names.
- no speculative TripPhotograph/TripNote stores were added.
- `GetAsync` is a direct key read; no repeat of the #166 GetAll+filter anti-pattern.
- local active-trip invariant is enforced inside the readwrite transaction; same-trip replay is allowed, second distinct active trip for the same owner is rejected.
- ownership reassignment by Trip Id is rejected and all reads remain owner-scoped.
- active Trip recovery is persisted, not held in volatile UI state.
- local Trip JSON preserves status, timestamps, place, location provenance/privacy, sync status and SyncedAt with the expected camel-case enum contract.
- current Catch cleanup cannot touch the new Trip store; Trip retention remains correctly deferred to T4.
- scope remains T2-only: no UI, sync, Catch.TripId, photos, notes or timeline.
- build-test is green on head `443e3001978ddf635fcee55b6b0b4cc8ac6de3b4`; browser-e2e workflow is skipped as expected for this slice, while the PR reports the dedicated browser storage suite green.

The v4->v5 upgrade is the highest-risk part of this slice, and the real-browser upgrade coverage with seeded Catch metadata/blob survival is the right test boundary.

Review verdict: ready from my side.

PR #184, opened 2026-08-27
FLB#175 | Trip synchroniser and logbook coordinator

### LINKED ISSUE #175: Trips T4 | Trip synchroniser and logbook coordinator

Part of #108. Depends on #172 (T1) and #173 (T2).

## Goal

Trips created offline synchronise safely on reconnect, in an order that guarantees dependent catches and media never reference a trip the server does not have.

## Scope

- `ITripSynchroniser` in `Features/Trips/Offline/Synchronisers/`, mirroring `CatchSynchroniser`: metadata scan for pending work, idempotent upsert, typed retryable vs non-retryable classification, interrupted-state recovery.
- A **thin** `ILogbookSynchroniser` coordinator — one method, no business logic — that runs **trips, then catches**, and raises a single `StateChanged`.
- `MainLayout` and `OfflineReconnectService` call the coordinator instead of `ICatchSynchroniser` directly.
- Cleanup rules: **active and pending trips are never evicted**; a completed, fully-synced trip becomes eligible after the same 24-hour window as catches, keyed on `SyncedAt`.
- Deterministic server-side reconciliation for two independently-created offline active trips — **the implementation detail is decided and reviewed in this slice** per #108.

## Explicitly not

Do **not** widen `ICatchSynchroniser` into a combined service. `CatchSynchroniser` already carries retry classification, interrupted-state recovery, missing-server-record recovery and cache cleanup; folding trips in produces an unreviewable class.

## Acceptance criteria

- [ ] Given a trip created offline, when connectivity returns, then it upserts to the server.
- [ ] Given a trip and catches created offline, when synchronisation runs, then the trip upserts **before** any catch referencing it.
- [ ] Given the same trip synchronised twice, when the second attempt runs, then no duplicate is created and no error is surfaced.
- [ ] Given a transient failure, when synchronisation retries, then it succeeds without user intervention and local data is retained throughout.
- [ ] Given a non-retryable failure, when it is classified, then it does not retry indefinitely on every login and no local data is discarded.
- [ ] Given a synchronisation pass, when it completes or fails, then `StateChanged` is raised once.
- [ ] Given an active trip, when cleanup runs, then it is never evicted regardless of age or sync state.
- [ ] Given a pending or failed trip, when cleanup runs, then it is never evicted.
- [ ] Given a completed, fully-synced trip older than the retention window, when cleanup runs, then it becomes eligible.
- [ ] Given two active trips created independently offline, when both sync, then reconciliation is deterministic and loses no data.
- [ ] Existing catch synchronisation and its offline-created-catch reconnect journey remain unregressed.

## Test requirements

- Follow the existing `CatchSynchroniserTests` patterns, including counter/spy assertions proving work was **not** performed.
- Prove ordering explicitly, not incidentally.
- Prove the cleanup exclusions for active and pending records.
- Prove `StateChanged` count on both success and failure paths.

## Definition of done

kingkeamo:
Closes #175
Part of #108

T4 of the Fishing Trips slices. Trips now reach the server, come back reconciled, and age out of the local cache.

## What this implements

**`TripSynchroniser`** (`Features/Trips/Offline/Synchronisers/`)
- Scans pending trips for the signed-in owner, sends each through `ITripClient` to `POST api/trips`, applies the server's returned lifecycle (`Status`, `EndedOn`) back onto the local record, and marks it `Synchronised` with a `SyncedAt` stamp.
- Failures become `FailedToSynchronise` — still pending, so the next run retries.
- A per-trip in-flight guard stops the same trip being sent twice concurrently.
- Local state stays authoritative: after the server responds, the local record is re-read and the response is only applied if the trip's synchronised content is unchanged. A trip edited or deleted mid-flight is never overwritten by the older response, and never resurrected.

**`LogbookSynchroniser`** (`Common/Offline/Synchronisers/`)
- Thin coordinator mirroring `ICatchSynchroniser`'s method shape, so call sites swapped one-for-one. Runs trips before catches under a no-concurrent-runs lock.
- `MainLayout` and `OfflineReconnectService` now call the coordinator rather than the Catch synchroniser directly. The reconnect owner-verification boundary is unchanged.

**Local cache retention**
- Synced *completed* trips older than 24 hours are evicted locally. Active trips are never evicted however old, and unsynced trips are never evicted regardless of age.
- The cleanup transaction is scoped to the `trips` store only — it cannot reach catches or photographs.

## Deliberate behaviour change: reconciliation replaces the 409

Two devices offline can both start an Active trip. T1 rejected the second with a 409. That leaves the angler with data the server refuses and no way to resolve it.

The second Active trip is now reconciled **inside the persistence transaction**:

- the later `StartedOn` stays Active;
- the earlier becomes Completed with `EndedOn` = the winner's start time;
- ties on `StartedOn` break on `Id`;
- **both records are kept** — nothing is discarded;
- the outcome is identical whichever order they arrive in.

Because this happens in the same transaction as the upsert, the partial unique index `ON "Trip" ("OwnerUserId") WHERE "Status" = 'Active'` is never transiently violated. **The index is unchanged** and still guards the genuinely concurrent race two transactions can't see; that case now converges on the next synchronisation run instead of retrying forever.

This required removing the `TripService` active pre-check (the check is no longer correct — reconciliation is not rejection) and the then-unused `ITripRepository.GetActiveAsync`, and retargeting four existing test files.

## Not in this slice

- **A trip sync failure does not block catch sync.** No catch references a trip until T5, so the coordinator establishes and tests the ordering only. Cleanup *does* propagate: trip cleanup failure stops catch cleanup.
- No UI changes — this slice is synchronisation only.

## Tests

| Suite | Result |
|---|---|
| `dotnet test FishingLogBook.sln` | 1846 passed |
| `npm test` | 294 passed (27 files) |
| `npm run test:browser` | 116 passed, 6 skipped |

New coverage:
- **Synchroniser (22)** — unknown owner, nothing pending, offline, another angler's trip, server rejection, conflict settling on the next run, retry after failure, concurrent double-send, local change mid-flight, local delete mid-flight, server lifecycle applied, batch invalidation, no re-send after success, full payload.
- **Coordinator (9)** — cancellation propagates, trip failure still runs catches, no concurrent runs, owner resolution, `StateChanged`, trips-before-catches ordering, cleanup ordering and propagation.
- **Reconciliation against real PostgreSQL (8)** — earlier demoted, earlier arrival stored completed, order-independence, exactly one Active after four upserts, another angler untouched, replay idempotent, tie-break on id, `EndedOn` never before `StartedOn`.
- **`TripClient` HTTP contract (4)** — exact URI and verb, body, deserialised lifecycle, failure on 4xx/5xx.
- **IndexedDB interop (3)** and **JS store (14)** — owner guards, cursor filtering, retention eligibility, numeric/string sync status.
- **Browser (6 × 2 engines)** — pending scan and retention cleanup against real IndexedDB, surviving reload, catches untouched.

None. No schema change, no Terraform, no manual post-merge steps.

## Self review

```
- Issue/AC: PASS
- Architecture/CQRS: PASS — reconciliation sits at the persistence boundary, not the handler
- Rules: PASS
- Security/privacy: PASS — owner-scoped at store, synchroniser and cleanup; diagnostics carry only tripId + errorType, never coordinates, payloads or tokens
- Database/migrations: N/A — no schema change; partial unique index unchanged
- Tests/coverage: PASS
- Browser: PASS — trip-sync-cache.spec.js proves the pending/cleanup cursors against real IndexedDB
- Web/UI/localisation: N/A — no user-visible copy in this slice
- Code smells: PASS
- CI/deployment: PASS
```

1. `OutranksActive` tie-break (identical `StartedOn`) had no test — the id-comparison branch was unproven.
2. The 409 path survives for the concurrent race, and the AC requires proving it does not retry endlessly.
3. Reconnect/MainLayout/Landing test bases were still bound to `ICatchSynchroniser`, leaving the coordinator swap unproven at those call sites.

Fixes made:
1. Added `ItShouldBreakATieOnTheStartTimeDeterministically` against real PostgreSQL.
2. Added `ItShouldSendAConflictedTripOnlyOncePerRunAndSettleOnTheNextRun` — one send per run, settled on the next.
3. Retargeted the three test bases to `ILogbookSynchroniser`, preserving every existing `DidNotReceive()` / `Received(1)` so the owner-verification boundary stays proven.

Remaining deliberate gaps:
- No browser test drives the C# synchroniser end to end; the Playwright host is not an authenticated Blazor host (per `self-review.md`). The JS storage layer it depends on is covered in a real browser, and the orchestration is covered at unit level. No follow-up testing-infrastructure ticket is warranted yet.
- Trip sync failure not blocking catch sync is a scoped T4 decision — revisit at T5 when catches reference trips.


kingkeamo (commented):
Reviewed T4 against #175/#108 and the actual changed code paths. I found no code blocker in the synchroniser/coordinator/reconciliation design. The focused TripSynchroniser, thin LogbookSynchroniser ordering, reconnect integration, stale-completion protection, active-trip reconciliation, and 24h completed-trip retention are all consistent with the agreed architecture. One deliberate behaviour to keep in mind for T5: trip sync failure currently does not block unrelated catch sync because catches do not yet reference Trip; revisit dependency-sensitive catch blocking once Catch.TripId exists. CI build-test is still in progress on d161692ac71d28ce330e84a4159a223c608e0fd4, so merge only after that is green. Ready from my side subject to CI.

PR #186, opened 2026-08-27
FLB#177 | Trip photographs

### LINKED ISSUE #177: Trips T6 | Trip photographs

Part of #108. Depends on #174 (T3) and #175 (T4).

## Goal

An angler adds photographs that belong to the day rather than to a catch — sunrise, the boat, lunch, the weather — and they persist and sync offline-first.

## Scope

- `TripPhotograph` domain type, `"TripPhotograph"` table (FK to `"Trip"`, index on `TripId`), object keys `trips/{userId}/{tripId}/{photoId}`.
- `tripPhotographs` IndexedDB store, blobs keyed by photograph id with the id list carried in the trip's metadata JSON — the #166 pattern, so a single trip read stays cheap.
- API: upload-url → PUT presigned → record, plus delete, mirroring the catch photograph endpoints and reusing the same presign service.
- Add Photo action on the active trip; carousel display on the trip.
- Sync via `ITripSynchroniser`, including the missing-server-record recovery pattern from #166 for a 404 on upload-url.

## Reuse, do not rebuild

`PhotographPicker`, `PhotographCarousel`, `PhotographDetails`, `IPhotographPreparationService` and `IPhotographMetadataService` are already catch-agnostic — `PrepareAsync(file, source, ct)` has no `CatchId`. Reuse them unchanged, including sanitisation and fail-closed behaviour.

**A trip photograph is never a `CatchPhotograph` with a synthetic `CatchId`.** That would break `FkCatchPhotographCatch` and corrupt every "photographs of this catch" query.

## Capture time

Use the trustworthy capture time when available — `PhotographMetadataModel.HasTrustworthyCapturedOn`, true only for `ExifOriginal`/`ExifDigitized`. Otherwise record the time it was added to the trip. **Never** substitute a file timestamp as trustworthy EXIF. Persist both values distinctly so the timeline (T8) can order honestly.

Photo EXIF **location** for trip media remains deferred; the pipeline continues to sanitise coordinates out of the stored bytes.

## Out of scope

Video · timeline rendering (T8) · notes (T7) · photo-derived trip location.

## Acceptance criteria

- [ ] Given an active trip, when a photograph is added from the camera, then it persists locally and appears on the trip.
- [ ] Given an active trip, when a photograph is added from the gallery, then the same holds even when the platform stripped all metadata.
- [ ] Given a photograph with trustworthy EXIF capture time, when stored, then that time is persisted as the capture time.
- [ ] Given a photograph with no trustworthy capture time, when stored, then the added-to-trip time is used and no file timestamp is treated as EXIF.
- [ ] Given trip photographs offline, when connectivity returns, then each uploads and is recorded exactly once.
- [ ] Given a trip metadata/list read, when it runs, then no photograph blob is read.
- [ ] Given a trip photograph is deleted, when the trip is read, then it no longer appears and the stored blob is removed.
- [ ] Given a photograph that fails sanitisation, when it is added, then it is rejected and nothing is persisted.
- [ ] Catch photographs and their table are entirely unaffected.

## Test requirements

- bUnit for the add/display flow; HTTP-contract tests for the new client methods proving exact URI, verb, body and failure behaviour.
- Prove the upload sequence end to end with consistent ids, object keys and content types.
- Browser test proving trip metadata reads never touch the blob store.
- Prove the stripped-metadata path explicitly — do not assume metadata is present.

## Definition of done

- [ ] EN and FR resources added

kingkeamo:
Closes #177
Part of #108

T6 of the Fishing Trips slices. An angler can photograph the day itself — the sunrise, the boat, the weather — not just the fish.

## Separate ownership

`TripPhotograph` is its own domain type and its own table, with a foreign key to `"Trip"`, an index on `TripId`, and a unique index on the object key. **A trip photograph is never a `CatchPhotograph` with a borrowed `CatchId`** — that would break `FkCatchPhotographCatch` and corrupt every "photographs of this catch" query. A repository test asserts no catch photograph row and no catch row is created, and a grep-level check confirms `CatchId` appears nowhere in the trip photograph path.

```sql
CREATE TABLE IF NOT EXISTS "TripPhotograph"
(
    "Id" uuid NOT NULL, "TripId" uuid NOT NULL, "ObjectKey" text NOT NULL,
    "ContentType" text NOT NULL, "CapturedOn" timestamptz NULL, "AddedOn" timestamptz NOT NULL,
    "CreatedOn" timestamptz NOT NULL DEFAULT now(), "UpdatedOn" timestamptz NOT NULL DEFAULT now(),
    CONSTRAINT "PkTripPhotograph" PRIMARY KEY ("Id"),
    CONSTRAINT "FkTripPhotographTrip" FOREIGN KEY ("TripId") REFERENCES "Trip" ("Id")
);
```

No `ON DELETE` clause, so the default no-action applies: a trip cannot be deleted out from under its photographs, and nothing can cascade into catch data. Trip deletion is not a V1 feature; when it arrives it gets its own deliberate semantics.

There is no `OwnerUserId` column — ownership comes from the trip, and every operation verifies it there.

Object keys are `trips/{userId}/{tripId}/{photoId}`, asserted to start `trips/` and never `catches/`.

## Reuse, not rebuild

`PhotographPicker`, `PhotographCarousel` and `IPhotographPreparationService` were already catch-agnostic, so they are used unchanged. Trips inherit sanitisation, fail-closed rejection, and camera-or-gallery selection for free. **There is exactly one preparation path**: no EXIF parsing or JPEG sanitisation code exists anywhere under `Features/Trips`, and only sanitised bytes ever reach IndexedDB, the API or R2. A photograph that fails preparation is rejected and nothing is persisted.

Each photograph owns its own metadata independently — a camera photograph with nothing in it cannot suppress a gallery photograph's EXIF, proven by a test that adds one of each and checks both.

## Capture time, stated exactly

- `CapturedOn` is stored **only** when the metadata is genuinely trustworthy (`ExifOriginal` / `ExifDigitized`).
- A file timestamp is **never** promoted to a capture time — tested explicitly with `FileLastModified`.
- `AddedOn` is always recorded, and both values persist distinctly so T8's timeline can order honestly.
- Missing metadata is expected, not exceptional: Google Photos strips it inconsistently, and the stripped path is tested directly rather than assumed.

**Photo GPS is not stored in T6 at all.** Not on the photograph, not derived into `Trip.Location`, not borrowed between photographs, not applied to catches, never rendered, never logged. The pipeline keeps sanitising coordinates out of the stored bytes. No product behaviour depends on photo location, so retaining it would have been speculative surface.

## Local-first persistence

Photographs persist the instant they are chosen, with no network involved. Blobs live in a new `tripPhotographs` store keyed by photograph id; their metadata rides on the trip record — the #166 pattern — so **reading a trip never opens the blob store**. That is proven in a real browser by patching `IDBDatabase.transaction` and asserting `tripPhotographs` is never among the stores opened during trip metadata, active-trip and pending reads.

The database moves **v5 → v6, additively**: one new store, nothing renamed, nothing rewritten, no backfill. A browser test seeds a realistic v5 database — a catch with a 64KB blob and a `TripId`, plus a trip — opens it through the app, and asserts the version, the exact four-store set, and that every seeded record and blob survives.

## Synchronisation ordering

A focused `ITripPhotographSynchroniser` handles presign → PUT → record. I deliberately did **not** widen `TripSynchroniser`: upload, retry and stale-write handling is enough complexity to deserve its own service, and it only ever asks a read-only question about trips, so there is no cycle.

Coordinator order is now **trip metadata → trip photographs → catches**, because a photograph cannot be recorded before its trip exists server-side.

**The exact rule:** a photograph waits until `IsTripReadyForServerAsync` reports its trip is on the server. Only that photograph waits — photographs of other, ready trips upload in the same run. A photograph that fails to upload keeps its sanitised bytes, stays pending, retries on the next run, and **never** marks the trip failed or blocks unrelated catch synchronisation.

Stale-write protection: bytes are re-read after recording, and if the photograph was removed locally mid-flight nothing is written back — a completed upload cannot resurrect a deleted photograph. A per-photograph in-flight guard prevents double upload.

## Retention and dependency protection

`GetTripsAwaitingCatchesAsync` becomes `GetTripsAwaitingDependentsAsync` and now also holds back any trip with a pending photograph, so retention cannot strand unsynchronised media. The check reads metadata only — no photograph bytes. When a trip finally is evicted, its photograph blobs are deleted in the same transaction rather than orphaned.

## Tests

| Suite | Result |
|---|---|
| `dotnet test FishingLogBook.sln` | 1996 passed |
| `npm test` | 325 passed (31 files) |
| `npm run test:browser` | 140 passed, 6 skipped |

76 new tests: repository 13 (real PostgreSQL — FK rejection, no cross-trip move, capture/added round-trip, trip-prefixed key, idempotent replay, no catch contamination, timeline order, reconciliation survival, scoped delete) · service 16 · API 8 · synchroniser 11 · dependency 4 · coordinator 3 · client HTTP contract 7 · JS↔.NET interop 11 · Active Trip UI 15 · vitest 19 · browser 6 × 2 engines.

One additive migration, applied by the normal DbUp run. No Terraform, no backfill, no other manual post-merge steps.

## Self review

```
- Issue/AC: PASS
- Architecture/CQRS: PASS — endpoint → mediator → handler → ITripPhotographService → repositories; the synchroniser only asks a read-only question
- Rules: PASS
- Security/privacy: PASS — trip ownership verified server-side, identical error for unknown and foreign trips, no coordinates stored, rendered or logged
- Database/migrations: PASS — additive, no cascade, no backfill
- Tests/coverage: PASS
- Browser: PASS — v5→v6 upgrade, blob granularity, owner scoping, reload recovery, retention dependency
- Web/UI/localisation: PASS — EN and FR
- Code smells: PASS
- CI/deployment: PASS
```

Findings discovered during self-review — two would have lost data:

1. **`putTrip` wiped the photograph list.** `ActiveTripService.FinishAsync` saves a `TripModel` snapshot captured before any photograph was added, so finishing a trip would have erased every photograph's metadata and orphaned its blobs.
2. **Retention orphaned blobs.** Evicting a trip removed the trip record but left its photograph blobs in `tripPhotographs` permanently.
3. An empty Mapster `NewConfig` with no `.Map` calls, which `cqrs.md` explicitly says to delete.
4. A global `COUNT(*)` assertion polluted by other suites sharing the Postgres fixture.

Fixes made:

1. `putTrip` no longer owns the photograph list — it carries forward whatever is stored. Only the photograph store mutates it. Five vitest cases cover finishing from a stale snapshot, a sync write omitting the field, and a genuinely new trip.
2. Cleanup deletes the evicted trip's photograph blobs in the same transaction, with tests for both the evicted and the retained case.
3. Deleted the empty mapping; convention already covers the pair.
4. Rescoped the assertion to the data under test.

Also checked and clear: no raw bytes stored before sanitisation, no duplicate preparation path, no fake `CatchId`, no object-key collision, no metadata stripped by our own code, no camera photograph suppressing gallery metadata, no blob-heavy list reads, owner scope on every operation, no photograph uploaded before its trip, no photograph failure blocking catches, no stale completion resurrecting a deleted photograph, no pending photograph evicted, no destructive upgrade, no exact GPS rendered or logged, and no T7/T8 scope.

Remaining deliberate gaps:

- **No list endpoint.** Nothing in T6 reads trip photographs from the server — the UI is local-first and T8 owns the read model. Adding one now would ship unused code.
- **Removing a *synchronised* photograph needs connectivity**, because it deletes the stored object before the record. Offline it shows a recoverable message and keeps the photograph rather than leaving an orphaned R2 object. Unsynchronised photographs — the common case on an active trip — remove offline. I did not invent a tombstone workflow for the other case.
- **No credentialed E2E.** `self-review.md` forbids test-only authentication, compile symbols or dual-host hacks to force an authenticated Blazor host, and production auth is Cognito/OIDC. The whole local journey — add, reload, recover, retention — is proven in a real browser on Chromium and WebKit; the credentialed server round trip stays a testing-infrastructure ticket and the full offline device journey remains T9.
- Physical-device acceptance is not claimed. The feature is built so that missing metadata is normal and selection still succeeds.


kingkeamo (commented):
Reviewed the actual T6 implementation and relevant tests. No code blocker found from this review. The separation between TripPhotograph and CatchPhotograph is clean, the v5→v6 IndexedDB upgrade is additive, the shared photograph preparation/sanitisation path is reused rather than duplicated, trip-photo sync is kept focused and ordered after trip metadata but before catches, and parent-trip retention is extended so pending trip media cannot be stranded. The stale-delete protection and the fix preventing stale Trip saves from wiping photograph metadata are especially important and look correct. Ready from my side once the build-test workflow is green.

PR #187, opened 2026-08-27
FLB#178 | Timestamped trip notes

### LINKED ISSUE #178: Trips T7 | Timestamped trip notes

Part of #108. Depends on #174 (T3) and #175 (T4).

## Goal

An angler jots quick notes during the day and one final note when finishing, and each note remembers when it was written so it can appear in the timeline at the right moment.

## Why timestamped rather than one trip note

Settled in #108: a single overall note has no time, so it cannot appear in the timeline at all — which quietly demotes the feature the product cares most about. Multiple timestamped notes cost one table with four columns, and they make the final "note about the day" just another note rather than a special case.

## Scope

- `TripNote` domain type; `"TripNote"` table (FK to `"Trip"`, index on `TripId`).
- `tripNotes` IndexedDB store; note ids carried in the trip's metadata JSON.
- `POST /api/trips/{tripId}/notes`, idempotent on a client-generated note GUID.
- Add Note action on the active trip; optional final note on the finish recap, persisted as an ordinary note.
- Sync via `ITripSynchroniser`.
- EN + FR localisation.

## Deliberately minimal

Plain text only. No rich text, no formatting toolbar, no attachments. One `MudTextField` with a length cap consistent with `CatchDetailConstants`.

## Out of scope

Timeline rendering (T8) · note editing or deletion unless trivially free · catch notes, which already exist and are untouched.

## Acceptance criteria

- [ ] Given an active trip, when a note is added, then it persists locally with its own creation timestamp.
- [ ] Given two notes added at different times, when the trip is read, then both are present with distinct timestamps in creation order.
- [ ] Given the finish recap, when a final note is entered, then it persists as an ordinary trip note.
- [ ] Given the finish recap, when no note is entered, then the trip finishes normally.
- [ ] Given an empty or whitespace-only note, when saving is attempted, then nothing is persisted.
- [ ] Given a note longer than the cap, when saving is attempted, then it is rejected per existing validation conventions.
- [ ] Given notes added offline, when connectivity returns, then each syncs exactly once.
- [ ] Given the same note synchronised twice, when the second attempt runs, then no duplicate is created.
- [ ] Catch notes are unaffected.

## Test requirements

- Validator tests: empty, whitespace, at the cap, one over the cap.
- bUnit for the add-note flow and the finish-with-note flow, asserting the exact model passed to the store.
- Idempotent replay proven at the repository level.

## Definition of done

- [ ] EN and FR resources added

kingkeamo:
Closes #178
Part of #108

T7 of the Fishing Trips slices. An angler jots quick notes through the day — *water dropped about a foot*, *fish rising near the reeds*, *changed to olive nymph* — and each one remembers when it was written.

## A separate, timestamped model

`TripNote` is its own domain type and its own table. It is **not** `Catch.Notes` reused, **not** a giant string on `Trip`, and **not** a persisted generic event row.

```sql
CREATE TABLE IF NOT EXISTS "TripNote"
(
    "Id" uuid NOT NULL, "TripId" uuid NOT NULL, "Text" text NOT NULL,
    "RecordedOn" timestamptz NOT NULL,
    "CreatedOn" timestamptz NOT NULL DEFAULT now(), "UpdatedOn" timestamptz NOT NULL DEFAULT now(),
    CONSTRAINT "PkTripNote" PRIMARY KEY ("Id"),
    CONSTRAINT "FkTripNoteTrip" FOREIGN KEY ("TripId") REFERENCES "Trip" ("Id")
);
CREATE INDEX IF NOT EXISTS "IxTripNoteTripId" ON "TripNote" ("TripId");
CREATE INDEX IF NOT EXISTS "IxTripNoteTripRecordedOn" ON "TripNote" ("TripId", "RecordedOn");
```

Four columns plus timestamps. No `OwnerUserId` duplication — ownership comes through the trip, and every operation verifies it there. No cascade, so nothing can reach catch data. Catch notes and their table are entirely untouched: no file under `Features/Catch` or `Application/Catches` changed.

`RecordedOn` is the timeline key, generated on the device at the moment of writing and carried unchanged through persistence and sync. **The server never substitutes its own insert time** — a repository test asserts `RecordedOn` differs from the row's own `CreatedOn`, so that can't quietly regress. Notes render in the reader's local time through the existing `ITimeService`; no new time abstraction.

## Deliberately lightweight

Plain text. Trimmed. Blank and whitespace-only rejected. Capped at `TripConstants.MaxNoteTextLength`, which is defined as `CatchDetailConstants.MaxNotesLength` so the two conventions cannot drift apart. No rich text, no formatting, no attachments, no mentions.

## Active Trip entry

Inline, not a modal — outdoors, taps matter:

```
[ Add note ]        →        What happened?
                             [ textarea            ]
                             [ Add note ] [ Cancel ]
```

Saving clears the editor, shows the note immediately with its local time, and keeps notes **oldest first** so the day reads forward — the same direction T8's timeline will run. The labelled textarea is keyboard reachable and the remove control carries an accessible label.

The same component appears on the finish recap, so #178's "final note about the day" is an **ordinary note**, not a special case. Entering nothing there finishes the trip exactly as before.

## Offline-first, and one owner of truth

Notes persist the instant they are written, with no network. They live on the trip metadata record, which means:

- **one authoritative metadata path** — no competing note lists;
- **no IndexedDB version change** (still v6) and no migration, because a note is metadata with no blob. T6 needed a separate store only to hold photograph *bytes*; its metadata already rode on the trip;
- a single trip read gives a timeline everything it needs.

A browser test asserts the version is still 6 and that no `tripNotes` store exists, so this is pinned rather than incidental.

`putTrip` no longer owns the note list — it carries forward whatever is stored, the same protection T6 added for photographs. **Finishing a trip therefore cannot discard notes written after the page loaded**, which is exactly the failure this would otherwise have shipped with.

## Delete

In scope: the T6 delete pattern made it close to free. An unsynchronised note deletes locally and works offline. A synchronised note deletes on the server first, then locally; if that fails the note is kept and a recoverable message appears. **No tombstone workflow was invented** for offline deletion of an already-synced note.

## Synchronisation ordering

A focused `ITripNoteSynchroniser` — note sync is deliberately **not** inside `TripSynchroniser`. Coordinator order is now **trips → trip photographs → trip notes → catches**: the trip must exist before a note's foreign key can be written, while photographs and notes are independent siblings, so one deterministic order keeps it testable.

**The exact rule:** a note waits until `IsTripReadyForServerAsync` reports its trip is on the server. A note that cannot be sent keeps its text, stays pending, retries next run, and **never** fails the trip, blocks photographs, or blocks catches. Both directions are tested — a photograph failure still lets notes and catches run; a note failure still lets catches run.

Notes carry a client-generated GUID and the server upserts on it, so replay cannot duplicate. Stale-write protection follows T4/T6: the note is re-read before the completion is applied, so a note deleted mid-flight is never resurrected.

## Privacy

Notes can contain private things. Diagnostics carry **only** `TripId`, `NoteId` and the error type — never the text. Asserted at both the synchroniser and the component using a deliberately private-sounding string, checking the message *and* every metadata value.

## Retention

`GetTripsAwaitingDependentsAsync` now also holds back any trip with a pending note, alongside pending catches and pending photographs. The check is metadata-only — no blob reads. When a trip is finally evicted its notes go with it, so nothing is orphaned.

## Tests

| Suite | Result |
|---|---|
| `dotnet test FishingLogBook.sln` | 2081 passed |
| `npm test` | 341 passed (32 files) |
| `npm run test:browser` | 154 passed, 6 skipped |

61 new tests: repository 11 (real PostgreSQL — FK rejection, no cross-trip move, idempotent replay, instant round-trip, server-clock guard, at the cap, recorded order, owner isolation, reconciliation survival, scoped delete) · service 11 · **validator 8** (empty, whitespace, at the cap, one over) · API 9 · synchroniser 11 · coordinator 2 · dependency 4 · Active Trip UI 19 · vitest 16 · browser 7 × 2 engines.

One additive migration, applied by the normal DbUp run. No Terraform, no backfill, no IndexedDB migration, no other manual post-merge steps.

## Self review

```
- Issue/AC: PASS
- Architecture/CQRS: PASS — endpoint → mediator → handler → ITripNoteService → repositories
- Rules: PASS
- Security/privacy: PASS — trip ownership verified server-side, identical error for unknown and foreign trips, note text never logged
- Database/migrations: PASS — additive, no cascade, no backfill
- Tests/coverage: PASS
- Browser: PASS — version unchanged, reload, ordering, owner scoping, pending detection without blob reads, retention, finish
- Web/UI/localisation: PASS — EN and FR, labelled textarea, keyboard-reachable controls
- Code smells: PASS
- CI/deployment: PASS
```

1. **The notes component trusted a stale trip snapshot.** It read `Trip.Notes` from the passed-in model, but after finishing, the page hands `CompletedTripSummary` a `TripModel` captured before the notes were written — so the finish recap would have shown none of them. The data was safe thanks to the `putTrip` preservation; the display was not.
2. **The synchroniser re-scanned every trip for each note** to perform its stale-write check — the same read-amplification shape as #166.

Fixes made:

1. Notes now load from the store, which is authoritative, with a snapshot fallback if that read fails. Three tests cover it: reading the store rather than the snapshot, the finish recap from a stale snapshot, and staying usable when the store throws.
2. Added targeted `GetForTripAsync` / `GetAsync`; the check now reads one trip instead of scanning them all.

Also checked and clear: `Catch.Notes` not reused, no giant note field on `Trip`, no persisted event table, no note text in any log or diagnostic, one metadata owner, no destructive IndexedDB upgrade, note sync outside `TripSynchroniser`, a note failure not blocking catches, no trip cleanup while a note is pending, no stale completion resurrecting a deleted note, no owner leakage, no server timestamp replacing the user-event timestamp, notes surviving reload, and no T8 scope.

Remaining deliberate gaps:

- **No list endpoint.** Nothing in T7 reads notes from the server; the UI is local-first and T8 owns the read model.
- **No note editing**, per #178 — it would introduce sync-conflict semantics we do not need before the timeline exists.
- **No credentialed E2E.** The auth harness boundary is unchanged (`self-review.md` forbids test-only authentication to force an authenticated Blazor host, and production auth is Cognito/OIDC). The complete local journey — add, reload, multiple notes, pending detection, retention, finish — is proven in a real browser on Chromium and WebKit. T9 owns the full device journey.

### DIFF
--- a/src/FishingLogBook.Api/Endpoints/TripEndpoints.cs
+++ b/src/FishingLogBook.Api/Endpoints/TripEndpoints.cs
@@ -69,7 +69,110 @@ public static class TripEndpoints
             .Produces(StatusCodes.Status404NotFound)
             .Produces(StatusCodes.Status503ServiceUnavailable);
 
+        endpoints.MapPost("/api/trips/{tripId:guid}/notes", RecordNoteAsync)
+            .WithName("RecordTripNote")
+            .WithTags("Trips")
+            .RequireAuthorization()
+            .Produces<TripNoteDto>(StatusCodes.Status200OK)
+            .Produces(StatusCodes.Status400BadRequest)
+            .Produces(StatusCodes.Status401Unauthorized)
+            .Produces(StatusCodes.Status404NotFound)
+            .Produces(StatusCodes.Status503ServiceUnavailable);
+
+        endpoints.MapDelete("/api/trips/{tripId:guid}/notes/{noteId:guid}", DeleteNoteAsync)
+            .WithName("DeleteTripNote")
+            .WithTags("Trips")
+            .RequireAuthorization()
+            .Produces(StatusCodes.Status204NoContent)
+            .Produces(StatusCodes.Status400BadRequest)
+            .Produces(StatusCodes.Status401Unauthorized)
+            .Produces(StatusCodes.Status404NotFound)
+            .Produces(StatusCodes.Status503ServiceUnavailable);
+
         return endpoints;
+    }
+
+    private static async Task<IResult> RecordNoteAsync(
+        Guid tripId,
+        RecordTripNoteDto note,
+        IMediator mediator,
+        ICurrentUser currentUser,
+        CancellationToken cancellationToken)
+    {
+        if (!currentUser.IsResolved)
+        {
+            return Results.Unauthorized();
+        }
+
+        var response = await mediator.Send(
+            new RecordTripNoteCommand
+            {
+                TripId = tripId,
+                Note = note
+            },
+            cancellationToken);
+        if (response.ValidationErrors is { Count: > 0 })
+        {
+            return Results.BadRequest(response);
+        }
+
+        if (response.Error is TripNoteNotFoundError)
+        {
+            return Results.NotFound();
+        }
+
+        if (response.Error is TripNoteInvalidError)
+        {
+            return Results.BadRequest(response);
+        }
+
+        if (response.IsFailure)
+        {
+            return Results.Problem(
+                title: response.ErrorMessage,
+                statusCode: StatusCodes.Status503ServiceUnavailable);
+        }
+
+        return Results.Ok(response.Note);
+    }
+
+    private static async Task<IResult> DeleteNoteAsync(
+        Guid tripId,
+        Guid noteId,
+        IMediator mediator,
+        ICurrentUser currentUser,
+        CancellationToken cancellationToken)
+    {
+        if (!currentUser.IsResolved)
+        {
+            return Results.Unauthorized();
+        }
+
+        var response = await mediator.Send(
+            new DeleteTripNoteCommand
+            {
+                TripId = tripId,
+                NoteId = noteId
+            },
+            cancellationToken);
+        if (response.ValidationErrors is { Count: > 0 })
+        {
+            return Results.BadRequest(response);
+        }
+
+        if (response.Error is TripNoteNotFoundError)
+        {
+            return Results.NotFound();
+        }
+
+        if (response.IsFailure)
+        {
+            return Results.Problem(
+                title: response.ErrorMessage,
+                statusCode: StatusCodes.Status503ServiceUnavailable);
+        }
+
+        return Results.NoContent();
     }
 
     private static async Task<IResult> CreatePhotographUploadAsync(
--- a/src/FishingLogBook.Application/Args/DeleteTripNoteArgs.cs
+++ b/src/FishingLogBook.Application/Args/DeleteTripNoteArgs.cs
@@ -1,0 +1,8 @@
+namespace FishingLogBook.Application.Args;
+
+public sealed class DeleteTripNoteArgs
+{
+    public Guid TripId { get; init; }
+
+    public Guid NoteId { get; init; }
+}
--- a/src/FishingLogBook.Application/Args/RecordTripNoteArgs.cs
+++ b/src/FishingLogBook.Application/Args/RecordTripNoteArgs.cs
@@ -1,0 +1,12 @@
+namespace FishingLogBook.Application.Args;
+
+public sealed class RecordTripNoteArgs
+{
+    public Guid TripId { get; init; }
+
+    public Guid NoteId { get; init; }
+
+    public string Text { get; init; } = string.Empty;
+
+    public DateTimeOffset RecordedOn { get; init; }
+}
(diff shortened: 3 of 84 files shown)


kingkeamo:
SHOULD FIX before merge: T7 still has no author attribution on `TripNote`. We just agreed this capability should be future-proof for shared trips via `CreatedByUserId` (or equivalent). Right now the domain model, DB table, repository and DTO path only carry `Id` / `TripId` / `Text` / `RecordedOn`, so adding shared-trip notes later would require a schema migration/backfill rather than simply enabling participant permissions/UI.

Please add `CreatedByUserId` now, set it from the trusted current user server-side (never from the client), persist and round-trip it, and keep current solo-trip behaviour unchanged. No shared-trip permission logic belongs in T7.

